PR #50, opened 2022-09-23
✨ Api add cors support

deleteLater:
enable cors for api


deleteLater:
maybe consider allow any origin later

PR #68, opened 2022-10-10
✨ feature flag part-2

deleteLater:
todo apis:

1. feature flag
- [x] update settings
- [x] update variations
- [x] check if key has been used
- [x] un-archive flag
- [x] update targeting
- [x] copy flag to another env
- [ ] tag tree

2. triggers
- [x] get triggers by target id
- [x] create trigger
- [x] update trigger
- [x] reset trigger token
- [x] delete trigger
- [x] run trigger

3. segment
- [x] get flag references

refactoring:
- add service base interface


deleteLater:
tag tree will be moved to another pr.

PR #81, opened 2022-10-24
✨ Try interacting with the demo

cosmic-flood:
1. Added DinoGame demo docker image to docker-compose-dev.yml. The DinoGame demo will be added to the FeatBit running container.
image

2. In the Guide Drawer of front-end, when the user clicks on "Try interacting with the demo", a demo page will be opened with 
the environment secret of the current environment in its URL.
image

Related trello card: https://trello.com/c/lBObHjPm/763-docs-interactive-with-demo


modules/front-end/src/app/core/components/guide/guide.component.ts
```diff
@@ -11,6 +13,17 @@ export class GuideComponent {
   isVisible = true;
   @Output() guideDrawerClosed: EventEmitter<any> = new EventEmitter();
 
+  currentProjectEnv: IProjectEnv;
```
cosmos-explorer: no need to create this property

cosmos-explorer (approved):

PR #92, opened 2022-11-03
📖 🏗 🔧Enable access of interactive demo hosted in remote server

cosmic-flood:
"3. Try interactive with demo" button in "Quick start guide" is a dynamic link which is composed by :
- Demo's remote server URL (which should be configurable).
- Environment secret of current environment in Portal.
- Evaluation server url which demo can communicate with for its FeatBit SDK.

image

So:

1. I modified the code which recomposes the link of "3. Try interactive with demo" button.
2. I added configurable params in environment.ts and environment.prod.ts. Because demo's url and evaluation server are different and depend on how users construct their network.
3. I updated docker-compose-dev.yml file that we can pass two enviroment's params from docker-compose.

PS:
I also added the number before each "Get Started" item. This is feedback from testing users.
image

### DIFF
--- a/docker-compose-dev.yml
+++ b/docker-compose-dev.yml
@@ -8,6 +8,8 @@ services:
       dockerfile: ./Dockerfile
       args:
         API_URL: http://localhost:5000
+        DEMO_URL: http://localhost:5173
+        EVALUATION_URL: http://localhost:5100
     depends_on:
       - api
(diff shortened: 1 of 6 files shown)


cosmos-explorer:
This won't work, you need to modify dockerfile to set the two new env variables

cosmos-explorer (approved):

PR #100, opened 2022-11-09
✨ Feat/feature flag reporting

cosmos-explorer:
Added feature flag reporting feature


modules/front-end/src/app/features/safe/feature-flags/details/reporting/reporting.component.less
```diff
@@ -1,22 +1,65 @@
...
+
+      .ant-select-selector {
+        border-radius: 32px;
+        height: 40px;
+
+        .ant-select-selection-item {
+          display: flex;
+          align-items: center;
+        }
+      }
```
deleteLater: Can we make this style global?

cosmos-explorer: done

modules/back-end/src/Application/FeatureFlags/GetFeatureFlag.cs
```diff
@@ -7,6 +7,7 @@ public class GetFeatureFlag : IRequest<FeatureFlag>
     public Guid EnvId { get; set; }
 
     public string Key { get; set; }
+
```
deleteLater: meaningless change

modules/back-end/src/Application/FeatureFlags/GetStatsByVariation.cs
```diff
@@ -0,0 +1,72 @@
...
+        RuleFor(x => x.Filter.FeatureFlagKey)
+            .NotEmpty().WithErrorCode(ErrorCodes.FeatureFlagKeyIsRequired);
+
+        RuleFor(x => x.Filter.IntervalType)
+            .Must(IntervalType =>
+            {
+                return IntervalTypeEnum.Month == IntervalType || IntervalTypeEnum.Week == IntervalType ||
+                       IntervalTypeEnum.Day == IntervalType || IntervalTypeEnum.Hour == IntervalType ||
+                       IntervalTypeEnum.Minute == IntervalType;
+            }).WithErrorCode(ErrorCodes.IntervalTypeIsRequired);
```
deleteLater: https://docs.fluentvalidation.net/en/latest/built-in-validators.html#enum-validator
dumb code, use `IsInEnum`

modules/back-end/src/Application/FeatureFlags/StatsByVariationFilter.cs
```diff
@@ -0,0 +1,18 @@
...
+}
+
+public static class IntervalTypeEnum
+{
+    public static string Month => "MONTH";
+    public static string Week => "WEEK";
+    public static string Day => "DAY";
+    public static string Hour => "HOUR";
+    public static string Minute => "MINUTE";
+}
```
deleteLater: - use `const` instead of `static` for constant fields
- 'Enum' postfix is bad, remove it

modules/back-end/src/Application/FeatureFlags/GetFeatureFlag.cs
cosmos-explorer: done

modules/back-end/src/Application/FeatureFlags/GetStatsByVariation.cs
cosmos-explorer: IsInEnum is only for enum parameter, I don't think it's proper for IntervalType as it's a string? @deleteLater

modules/back-end/src/Application/FeatureFlags/StatsByVariationFilter.cs
cosmos-explorer: done

modules/back-end/src/Application/FeatureFlags/GetStatsByVariation.cs
deleteLater: ok. how about extract this code to a method in `IntervalType` called `public static bool IsDefined(string intervalType)`, then call the method here.

deleteLater: and i think the error code can change to `ErrorCodes.InvalidIntervalType`

modules/back-end/src/Application/Experiments/Enums.cs
```diff
@@ -1,6 +1,6 @@
 ﻿namespace Application.Experiments;
 
-public class ExperimentStatus
+public static class ExperimentStatus
```
deleteLater, line 3: `const` is better for it's fields

modules/back-end/src/Application/EndUsers/GetFeatureFlagEndUserList.cs
```diff
@@ -0,0 +1,75 @@
...
+    public GetFeatureFlagEndUserListValidator()
+    {
+        RuleFor(x => x.Filter.FeatureFlagKey)
+            .NotEmpty().WithErrorCode(ErrorCodes.FeatureFlagKeyIsRequired);
+
+        RuleFor(x => x.Filter.From)
+            .GreaterThan(0).WithErrorCode(ErrorCodes.StatsFromIsRequired);
+
+        RuleFor(x => x.Filter.To)
+            .GreaterThan(0).WithErrorCode(ErrorCodes.StatsToIsRequired);
```
deleteLater: not a big problem, but i think the error code should be `ErrorCodes.InvalidStatsFrom` and `ErrorCodes.InvalidStatsTo`.

deleteLater: same for `GetStatsByVariationValidator`

modules/back-end/src/Application/Services/IEndUserService.cs
```diff
@@ -6,6 +6,7 @@ namespace Application.Services;
 
 public interface IEndUserService : IService<EndUser>
 {
+    Task<IEnumerable<EndUser>> GetListByKeyIdsAsync(Guid envId, IEnumerable<string> keyIds);
```
deleteLater: this method is redundant, just use `FindManyAsync`

modules/back-end/src/Infrastructure/Experiments/ExperimentService.cs
```diff
@@ -49,8 +49,8 @@ public async Task ArchiveExperiment(Guid envId, Guid experimentId)
                 CustomEventTrackOption = (int)iteration.CustomEventTrackOption,
                 CustomEventSuccessCriteria = (int)iteration.CustomEventSuccessCriteria,
                 CustomEventUnit = iteration.CustomEventUnit,
-                StartExptTime = iteration.StartTime.ToString("yyyy-MM-ddTHH:mm:ss.ffffff"),
-                EndExptTime = iteration.EndTime?.ToString("yyyy-MM-ddTHH:mm:ss.ffffff")
+                StartExptTime = iteration.StartTime.Millisecond,
+                EndExptTime = iteration.EndTime.HasValue ? iteration.EndTime.Value.Millisecond : null
```
deleteLater: https://learn.microsoft.com/en-us/dotnet/api/system.datetime.millisecond?view=net-6.0

In this file, there are many places u call `DateTime.Millisecond`, but its result is not what u expected. This should cause a BUG.

modules/back-end/src/Domain/FeatureFlags/StatsByVariationParam.cs
```diff
@@ -0,0 +1,30 @@
+﻿namespace Domain.FeatureFlags;
+
+public class StatsByVariationParam
```
deleteLater, line 3: i think `Param` is a bad suffix. maybe we can chang to `Query` here

modules/back-end/src/Infrastructure/Experiments/OlapService.cs
```diff
@@ -14,21 +15,52 @@ public OlapService(HttpClient httpClient)
     {
         _httpClient = httpClient;
     }
-
-    public async Task<ExperimentIteration> GetExptIterationResultAsync(ExptIterationParam param)
+    
+    private async Task<T?> GetFeatureFlagStatusByVariation<T>(string endPoint, Object param)
```
deleteLater: - weird method name
- use `object` instead of `Object` as parameter
- use `PostAsJsonAsync` to simplify code. example: `await _httpClient.PostAsJsonAsync("request-url", the-object-u-want-post);`

ref: https://learn.microsoft.com/en-us/dotnet/api/system.net.http.json.httpclientjsonextensions.postasjsonasync?view=net-6.0#system-net-http-json-httpclientjsonextensions-postasjsonasync-1(system-net-http-httpclient-system-string-0-system-text-json-jsonserializeroptions-system-threading-cancellationtoken)

modules/back-end/src/Application/FeatureFlags/GetStatsByVariation.cs
deleteLater: and `IntervalTypeEnum` is not exists, this code cannot compile.

modules/back-end/src/Application/EndUsers/GetFeatureFlagEndUserList.cs
cosmos-explorer: Done

modules/back-end/src/Application/Experiments/Enums.cs
cosmos-explorer: done

modules/back-end/src/Application/Services/IEndUserService.cs
cosmos-explorer: method removed

modules/back-end/src/Infrastructure/Experiments/OlapService.cs
cosmos-explorer: method removed

modules/back-end/src/Infrastructure/Experiments/ExperimentService.cs
cosmos-explorer: done

modules/back-end/src/Application/FeatureFlags/GetStatsByVariation.cs
cosmos-explorer: done
didn't find IntervalTypeEnum here

deleteLater (approved):

PR #130, opened 2022-11-30
✨ feat(ui): rename feature flag reporting tab to insights

cosmos-explorer:
image

### DIFF
--- a/modules/front-end/src/app/core/services/feature-flag.service.ts
+++ b/modules/front-end/src/app/core/services/feature-flag.service.ts
@@ -15,7 +15,7 @@ import {
   ISettingPayload,
   IVariationsPayload
 } from "@features/safe/feature-flags/types/details";
-import {IReportingFilter, IStatsByVariation} from "@features/safe/feature-flags/details/reporting/types";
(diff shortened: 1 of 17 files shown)


modules/front-end/src/app/core/services/feature-flag.service.ts
```diff
@@ -128,7 +128,7 @@ export class FeatureFlagService {
     return this.http.put<boolean>(url, flag.tags);
   }
 
-  public getReport(filter: IReportingFilter): Observable<IStatsByVariation[]> {
+  public getReport(filter: IInsightsFilter): Observable<IStatsByVariation[]> {
     const queryParam = {...filter};
     const url = `${this.baseUrl}/stats`;
     return this.http.get<IStatsByVariation[]>(url, {params: new HttpParams({fromObject: queryParam})});
```
deleteLater: Do we need to modify the api as well?

cosmos-explorer: done

deleteLater (approved):
NICE JOB!

PR #181, opened 2022-12-30
feat(ui): extract audit log list to a component

cosmos-explorer:
Extract audit log list to a component so it can be reused for all ref types


modules/front-end/src/app/core/components/audit-logs/audit-logs.component.ts
```diff
@@ -0,0 +1,142 @@
...
+
+@Component({
+  selector: 'audit-logs',
+  templateUrl: './audit-logs.component.html',
+  styleUrls: ['./audit-logs.component.less']
+})
+export class AuditLogsComponent implements OnInit, OnDestroy {
+  @Input() auditLogFilter: AuditLogListFilter;
+
+  private destory$: Subject<void> = new Subject();
```
deleteLater, line 20: why do we need this `destory$` subject

deleteLater (approved):

Issue #167, opened 2022-12-18
Able to add a custom attribute when create a new rule in Targeting tab


cosmic-flood:
### Discussed in https://github.com/orgs/featbit/discussions/139

<div type='discussions-op-text'>

<sup>Originally posted by **cosmic-flood** December  5, 2022</sup>
![image](https://user-images.githubusercontent.com/68597908/205584152-aa0a5378-6838-4c68-bf4b-2b99eb1136e4.png)

It'll be more convenient. </div>

PR #186, opened 2023-01-02
feat(ui): create user property when editing targeting rule condition

cosmos-explorer:
Isssue: https://github.com/featbit/featbit/issues/167
Make it possible to create a new user property when editiing rule condition on feature flag or segment targeting page

image


deleteLater (dismissed):

deleteLater (approved):
LGTM

Issue #188, opened 2023-01-03
[Bug]: Feature flag history tab content cannot scroll


cosmos-explorer:
### What happened?

Feature flag history tab content cannot scroll

### To Reproduce

1. Go to feature flag tageting page
2. Modify the feature flag multple times to have enough history items
3. Go to history tab
4. All history items are not visible if it has more than one page

### Expected behavior

We should be able to scroll page to see all items

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #189, opened 2023-01-04
🐛 fix: fixed the scroll bar not displaying bug on feature flag history tab

cosmos-explorer:
fix for https://github.com/featbit/featbit/issues/188


deleteLater (approved):
LGTM

PR #203, opened 2023-01-10
🐛 force to logout after reinitialized mongodb data

cosmos-explorer:
After regenerated data in mongodb, if we refresh the UI, we would see a blank page, that's because the org the current user loged in doesn't exist any more (id changed) . 

The current PR forces to logout the user in this situation.


deleteLater (approved):
LGTM

PR #214, opened 2023-01-13
✨ feat: use user keyId and feature flag key to generate splitting key

cosmos-explorer:

### DIFF
--- a/modules/evaluation-server/src/Domain/Core/UserVariation.cs
+++ b/modules/evaluation-server/src/Domain/Core/UserVariation.cs
@@ -1,4 +1,3 @@
-using System.Text;
 using System.Text.Json;
 
 namespace Domain.Core;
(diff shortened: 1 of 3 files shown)


modules/evaluation-server/src/Domain/Core/UserVariation.cs
```diff
@@ -78,7 +78,8 @@ public sealed class RolloutUserVariation : UserVariation
         }
         else
         {
-            var sendToExptKey = Convert.ToBase64String(Encoding.UTF8.GetBytes(userKeyId));
+            // create a new key to calculate the splitting percentage
+            var sendToExptKey = splittingKey + splittingKey;
```
deleteLater: still inefficient

cosmos-explorer:
Done in this PR #215

PR #215, opened 2023-01-13
✨ feat: rollout with custom key

cosmos-explorer:
allow user to set the key uppon which to dispatch end users
the default key is keyId


deleteLater (dismissed):
LGTM

deleteLater:
We need to sync this change to our server-side sdks before merge it @cosmos-explorer @dsun0720

deleteLater (approved):

PR #231, opened 2023-02-18
🏗 infra: change the docker network name

cosmos-explorer:
set the docker network to featbit-network-full to distinguish with the docker-compose-partial.yml


deleteLater (approved):
LGMT

PR #246, opened 2023-02-28
🏗CI: trigger ui change validations only when the UI code is changed

cosmos-explorer:
Only trigger ui change validations workflow when UI code is changed. I previously the workflow as required as not all changes are UI related


deleteLater (approved):
LGTM

PR #259, opened 2023-03-05
✨ Add description field to new feature form

sagar110599:

### DIFF
--- a/modules/front-end/src/app/features/safe/feature-flags/index/index.component.html
+++ b/modules/front-end/src/app/features/safe/feature-flags/index/index.component.html
@@ -210,11 +210,9 @@
         </nz-form-control>
       </nz-form-item>
       <nz-form-item>
-        <nz-form-label placeholder="A Small Description for feature flag">
-          Description
-        </nz-form-label>
-        <nz-form-control nzHasFeedback nzErrorTip="Cannot exceed 512 characters">
-          <textarea nz-input formControlName="description" placeholder="Description">
+        <nz-form-label i18n="@@common.description">Description</nz-form-label>
+        <nz-form-control nzHasFeedback nzErrorTip="Cannot exceed 512 characters" i18n-nzErrorTip="@@ff.description.cannot-exceed-512-characters">
+          <textarea nz-input formControlName="description" i18n-placeholder="@@common.description" placeholder="Description">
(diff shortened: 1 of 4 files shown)


modules/front-end/src/app/features/safe/feature-flags/index/index.component.html
```diff
@@ -209,6 +209,15 @@
...
       </nz-form-item>
+      <nz-form-item>
+        <nz-form-label placeholder="A Small Description for feature flag">
+          Description
+        </nz-form-label>
+        <nz-form-control nzHasFeedback nzErrorTip="Cannot exceed 512 characters">
+          <textarea nz-input formControlName="description" placeholder="Description">
+          </textarea>
+        </nz-form-control>
+      </nz-form-item>
```
cosmos-explorer, line 218: Great work, thank you @sagar110599 , may I ask you to do one more thing? We need to support i18n for all labels, we are using the official **@angular/localize** package. Can you modify this block of code like this?
```html
      <nz-form-item>
        <nz-form-label i18n="@@common.description">Description</nz-form-label>
        <nz-form-control nzHasFeedback nzErrorTip="Cannot exceed 512 characters" i18n-nzErrorTip="@@ff.description.cannot-exceed-512-characters">
          <textarea nz-input formControlName="description" i18n-placeholder="@@common.description" placeholder="Description">
          </textarea>
        </nz-form-control>
      </nz-form-item>
```

After that, please run: `npm run i18n`, this would generate the i18n resources, don't worry about the translation, we can do it afterwards.

sagar110599: Hi @cosmos-explorer , After updating the HTML as suggested and running the command npm run i18n following files got updated which are seen in screenshot. Do i need to commit those as well?? 
image

cosmos-explorer: Hi @sagar110599 , Yes, those files are the translation files, they need to be committed

sagar110599: Updated commit as per suggestions.

cosmos-explorer:
@all-contributors
please add @sagar110599  for code

PR #261, opened 2023-03-07
📖 doc: updated contribution task templates

cosmos-explorer:
Created two template files separately for API and UI


cosmos-explorer:
Aborted

PR #264, opened 2023-03-08
Added missing translations

cosmos-explorer:


modules/front-end/src/locale/messages.zh.xlf
```diff
@@ -1,5 +1,4 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<xliff version="1.2" xmlns="urn:oasis:names:tc:xliff:document:1.2">
+<?xml version="1.0" encoding="UTF-8"?><xliff version="1.2" xmlns="urn:oasis:names:tc:xliff:document:1.2">
```
deleteLater: Why this change?

cosmos-explorer: reverted

deleteLater (approved):
LGTM

PR #267, opened 2023-03-10
✨feat: added a tooltip for trigger url to tell how to invoke a trigger

cosmos-explorer:
It was not clear how to invoke a trigger, so I added this tooltip

image


deleteLater (approved):
LGTM

PR #309, opened 2023-03-23
🧹 refactor: replace id with key in feature flag urls

cosmos-explorer:
Replacing id with key in feature flag URLs makes them more user-friendly and easier for others to use.

### DIFF
--- a/modules/back-end/src/Application/FeatureFlags/ArchiveFeatureFlag.cs
+++ b/modules/back-end/src/Application/FeatureFlags/ArchiveFeatureFlag.cs
@@ -5,7 +5,9 @@ namespace Application.FeatureFlags;
 
 public class ArchiveFeatureFlag : IRequest<bool>
 {
-    public Guid Id { get; set; }
+    public Guid EnvId { get; set; }
+    
+    public string Key { get; set; }
 }
 
 public class ArchiveFeatureFlagHandler : IRequestHandler<ArchiveFeatureFlag, bool>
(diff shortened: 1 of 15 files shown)


modules/back-end/src/Application/FeatureFlags/ArchiveFeatureFlag.cs
```diff
@@ -29,7 +31,7 @@ public class ArchiveFeatureFlagHandler : IRequestHandler<ArchiveFeatureFlag, boo
 
     public async Task<bool> Handle(ArchiveFeatureFlag request, CancellationToken cancellationToken)
     {
-        var flag = await _service.GetAsync(request.Id);
+        var flag = await _service.FindOneAsync(x => x.EnvId == request.EnvId && x.Key == request.Key);
```
deleteLater: `FindOneAsync` may return `null`. Add a `null` check here and for a `null` flag, we should do nothing.

modules/back-end/src/Application/FeatureFlags/SetTags.cs
```diff
@@ -5,7 +5,8 @@ namespace Application.FeatureFlags;
 
 public class SetTags : IRequest<bool>
 {
-    public Guid Id { get; set; }
+    public Guid EnvId { get; set; }
+    public string Key { get; set; }
```
deleteLater: add blank line between props.

modules/front-end/src/app/core/components/expt-rules-drawer/expt-rules-drawer.component.ts
```diff
@@ -186,9 +186,9 @@ export class ExptRulesDrawerComponent {
   }
 
   doSubmit() {
-    const { id, targetUsers, rules, fallthrough, exptIncludeAllTargets } = this.featureFlag;
+    const { key, targetUsers, rules, fallthrough, exptIncludeAllTargets } = this.featureFlag;
 
-    this.featureFlagService.update({ id, targetUsers, rules, fallthrough, exptIncludeAllTargets })
+    this.featureFlagService.update({ key, targetUsers, rules, fallthrough, exptIncludeAllTargets })
```
deleteLater: rename `update` to `updateTargeting`

deleteLater (approved):
LGTM

PR #319, opened 2023-04-08
🐛 fix: audit log and flag changes not written or published when patching update feature flag

cosmos-explorer:
- create audit log
- publish flag changes


deleteLater (approved):
LGTM

PR #325, opened 2023-04-18
📖 doc: add help wanted issue template

cosmos-explorer:
This template has already been created before, this template is to add it as issue template so people can select it when creating an issue


deleteLater (approved):
LGTM

PR #364, opened 2023-05-15
✨ Create user if it does not exist when user login by email

cosmos-explorer:
For try it online environment, when a user log in

- create a new user if the email does not exist and log in
- log in if user exist


cosmos-explorer:
wrong PR, it is supposed for TIO repo not for the main repo

PR #377, opened 2023-05-20
✨feat: make it possible to copy each env secret of the current env

cosmos-explorer:
Screenshot

image


modules/front-end/src/app/core/components/header/header.component.ts
```diff
@@ -61,16 +67,20 @@ export class HeaderComponent implements OnInit {
 
     this.cannotReadProjectsMsg = $localize`You don't have permissions to read project list, please contact the admin to grant you the necessary permissions`;
     this.cannotReadEnvsMsg = this.canListProjects ? $localize`You don't have permissions to read environment list, please contact the admin to grant you the necessary permissions` : $localize`You don't have permissions to read project and environment list, please contact the admin to grant you the necessary permissions`;
```
deleteLater, line 69: Strange logic 🤔

deleteLater (approved):
LGTM

PR #378, opened 2023-05-23
🧹 refactor: replace `lodash` with `lodash-es`

defqd:
Imported specific methods individually instead of import the whole 'lodash' library


modules/front-end/angular.json
cosmos-explorer, line 1: Hello @defqd , Thank you for your PR.

Have you check the recommended way in this answer? https://stackoverflow.com/a/62589268
With this, we don't need to add **allowedCommonJsDependencies**

defqd: Hello @cosmos-explorer, I checked. The warning has disappeared.

modules/front-end/package.json
```diff
@@ -21,6 +21,7 @@
     "@angular/platform-browser-dynamic": "^14.0.0",
     "@angular/router": "^14.0.0",
     "@antv/g2": "^4.2.7",
+    "lodash-es": "^4.17.21",
     "lodash.differenceby": "^4.8.0",
```
cosmos-explorer: You can remove 
- lodash.differenceby
- lodash.get
- lodash.intersectionby
- lodash.keyby

Then rerun `npm i`

cosmos-explorer (approved):

cosmos-explorer:
LGTM! Thank you very much

PR #381, opened 2023-05-25
🚀Pro Manifest Improvements

wss-rbrennan:
Found that basic configurations are needed when deploying to non local cluster.

This PR adds the following

- cleanupInfrastructure.ps1 
  a script to quickly delete installed infrastructure components, 
- Add zookeeper-zoo-configmap.yaml
  a basic zoo.cfg file for single instance operation
- init containers to configure kafka and zookeeper
- setting security context to user 1001 to match bitnami containers
- set default resource requests and limits for all pods.  These are arbitrary and should be set to the users specific needs.

### DIFF
--- a/kubernetes/pro/application/ui-service.yaml
+++ b/kubernetes/pro/application/ui-service.yaml
@@ -3,7 +3,7 @@ kind: Service
 metadata:
   labels:
     app: ui
-  name: standard-ui
+  name: ui
 spec:
   ports:
     - name: "8081"
--- a/kubernetes/pro/application/ui-service.yaml
+++ b/kubernetes/pro/application/ui-service.yaml
@@ -3,7 +3,7 @@ kind: Service
 metadata:
   labels:
     app: ui
-  name: ui
+  name: standard-ui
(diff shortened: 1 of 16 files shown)


cosmos-explorer:
Hello @wss-rbrennan, when I test the code, I see this error, it cannot start the kafka pod, it may be related to this change:
`setting security context to user 1001 to match bitnami containers`
image

kubernetes/pro/application/ui-service.yaml
```diff
@@ -3,7 +3,7 @@ kind: Service
 metadata:
   labels:
     app: ui
-  name: ui
+  name: standard-ui
```
cosmos-explorer: Why this change?

wss-rbrennan: ui change was a mistake, fixed with 36a4f4f, security context fixed with 8688832

cosmos-explorer (approved):

cosmos-explorer:
LGTM, thank you

PR #406, opened 2023-06-20
🐛 fix: user can create duplicate key

deleteLater:
This pr fix this bug by:
- Added feature flag key validation before creating it in the API
- Enabled the "Save" button only after successful validation


cosmos-explorer (approved):
LGTM

PR #410, opened 2023-06-23
fix: data sync upload returns 415

benirave:
I have found an issue related to the upload of a JSON file using the _/data-sync/upload_ endpoint. In this case, the API consistently returns a _415 - Unsupported Media Type_ error.

The root cause of this issue lies within the **ApiControllerBase** class, where the Consume attribute is set to application/json as the content type. However, the application/json content type is not appropriate for the Upload endpoint, which requires multipart/form-data instead.

### DIFF
--- a/modules/back-end/src/Api/Controllers/DataSyncController.cs
+++ b/modules/back-end/src/Api/Controllers/DataSyncController.cs
@@ -27,6 +27,7 @@ public class DataSyncController : ApiControllerBase
     [RequestSizeLimit(510 * 1024 * 1024)]
     // single file max size: 500MB 
     [RequestFormLimits(ValueLengthLimit = 500 * 1024 * 1024)]
+    [Consumes("multipart/form-data")]
     public async Task<ApiResponse<bool>> UploadAsync(Guid envId, IFormFile? file)
     {
         if (file is not { Length: > 0 })


deleteLater:
Thanks @benirave ❤️ 

To solve this problem, I prefer to remove the `[Consumes(MediaTypeNames.Application.Json)]` attribute from `ApiControllerBase` since this attribute is not very meaningful. By removing the attribute, we can rely on the [request inference from ApiController attribute](https://learn.microsoft.com/en-us/aspnet/core/web-api/?view=aspnetcore-6.0#binding-source-parameter-inference-1), specifically [Multipart/form-data request inference](https://learn.microsoft.com/en-us/aspnet/core/web-api/?view=aspnetcore-6.0#multipartform-data-request-inference-1).

benirave:
That would be the cleanest solution for sure, I can make that change.

Fyi, if we remove the global "Consume" attribute, then by default the allowed content types would also include 'text/json' and 'application/*+json'. Is that alright with you?

deleteLater:
Yeah, I did notice that as well, and I think it's acceptable and won't break anything 🍀

benirave:
It most probably won't break anything. Although, if you prefer to restrict the allowed input content-types we can add the `ConsumeAttribute` to the MVC options:

`builder.Services.AddControllers(x =>
{
    x.Filters.Add(new ConsumesAttribute(MediaTypeNames.Application.Json));
})`

However, this would still allow `application/*+json`. To exclude all other content-types completely we could remove these from the SupportedMediaTypes.:

`    .AddMvcOptions(x =>
    {
        var jsonInputFormatter = x.InputFormatters.OfType<SystemTextJsonInputFormatter>().First();
        jsonInputFormatter.SupportedMediaTypes.Remove("text/json");
        jsonInputFormatter.SupportedMediaTypes.Remove("application/*+json");
    });
`

Btw, these default media types are coming from [SystemTextJsonInputFormatter](https://github.com/dotnet/aspnetcore/blob/213eb282fc6d9ce07a1e0a1fbe526c297694ebf6/src/Mvc/Mvc.Core/src/Formatters/SystemTextJsonInputFormatter.cs#L14).

Let me know if you would prefer to restrict the input content types.

deleteLater:
Thank you for your advice. I prefer to keep things simple, and at the moment, there is no obvious reason to restrict the input content types. 

If you have any further questions or suggestions, please feel free to share them.

deleteLater (approved):

PR #426, opened 2023-07-01
✨ feat: display variation names instead of valuess in insights page end user table

cosmos-explorer:
Before
image

After
image


deleteLater (approved):
LGTM

Issue #395, opened 2023-06-14
[Good First Issues]: Add a link to FeatBit docs website


cosmos-explorer:
# Description
Currently user can find the docs link only in the get started module, it's not always visible when user is using other modules. We would like to add a link of the docs in the top right of the page so it will be always visible. 

# Scope
- Service: UI
- Components: header
- Location:
  - modules/front-end/src/app/core/components/header
 
link target: https://docs.featbit.co
 
You can use this icon
``` 
<span nz-icon nzType="read" nzTheme="outline"></span>
```

# Screenshot
![image](https://github.com/featbit/featbit/assets/88151306/0472b19c-15da-4476-9e20-93fcfc84ea60)

# Setup dev environment

Please read the [doc](../blob/main/Development.md) to set up your development environment.

# Deadline

We usually give 2 weeks for a feature request, which means you have 2 weeks to implement the feature starting from the
day it is assigned to you. If you have any difficulties, please contact us
in [Slack](https://join.slack.com/t/featbit/shared_invite/zt-1ew5e2vbb-x6Apan1xZOaYMnFzqZkGNQ).

# How to implement

1. Fork the FeatBit repository.
2. Create a branch issues-{issues number}
3. Implement the task based on Description and Scope mentioned above.
4. Submit your PR

If the feature is UI related, make sure you complete the following tasks before submit your PR:

- merge featbit main branch into your working branch
- run `npm run i18n`
- commit **messages.xlf & messages.zh.xlf** file changes (if any)

# Contribution points

Each contribution counts a certain number of points, depending on its difficulty. Contributors earn the appropriate
points when the work is merged. Contribution points are used to describe the contributions that contributors have made.
They can also be used to receive rewards for community events. You can view the current community submissions on a
public [google sheet](https://docs.google.com/spreadsheets/d/1ukyXgi_jRPeXj7EAST0IrnPfLOQ6xDBkcyAJY9N-Yb4/edit#gid=1117970540).

Points: 4

# How to claim to solve the issue

If you want to implement this function, please leave a comment in this issue like:

> I'd like to implement this function, please assign this issue to me.

patel-aum:
Hi i can work upon this

deleteLater:
Thanks @patel-aum, I've assigned this issue to you 🍀

LeeBingler:
Hi, can i try ?

deleteLater:
Yeah @LeeBingler, feel free to create a PR :clap:

PR #433, opened 2023-07-05
✨ UI feat: add icon on header that redirect on the docs

LeeBingler:
Implement aspect of the header by adding a new button.
## Before:
beforeUI

## After :
afterUI

Issue #395

### DIFF
--- a/modules/front-end/src/app/core/components/header/header.component.html
+++ b/modules/front-end/src/app/core/components/header/header.component.html
@@ -7,7 +7,7 @@
 
   <div class="header-btns">
     <div class="header-btn feedback" (click)="openFeedbackModal()" i18n="@@common.feedback">Feedback</div>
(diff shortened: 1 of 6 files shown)


modules/front-end/src/app/core/components/header/header.component.html
```diff
@@ -7,6 +7,7 @@
 
   <div class="header-btns">
     <div class="header-btn feedback" (click)="openFeedbackModal()" i18n="@@common.feedback">Feedback</div>
+    <a class="header-btn doc-btn" href="https://docs.featbit.co"><span nz-icon nzType="read" nzTheme="outline"></span></a>
```
cosmos-explorer: Please add target="_blank" to the link so the doc page would be open in a new tab

cosmos-explorer:
@LeeBingler Thank you very much for your PR, please read my comment. 

Also, as mentioned in the issue description, you will need to merge the main branch into yours and run the command: npm run i18n before submit

PR #441, opened 2023-07-10
🧹 refactor: replace class-based resolvers with functional resolvers

cosmos-explorer:
Resolve is deprecated and will be removed in Angular 17, the prefered way is the ResolveFn


deleteLater (approved):
LGTM

PR #444, opened 2023-07-11
✨ feat: made UI to return 444 when seeing malicious scan requests

cosmos-explorer:
We keep seeing malicous scan requests from the internet, the PR return 444 for all unknown requests


deleteLater (approved):
LGTM

PR #451, opened 2023-07-14
🧹 tech-debt: replace class based CanActivate by function

cosmos-explorer:
WIP


modules/front-end/src/app/core/guards/iam.guard.ts
```diff
@@ -1,35 +1,21 @@
...
-    const isGranted = this.permissionsService.isGranted(generalResourceRNPattern.iam, permissionActions.CanManageIAM);
-
-    if (!isGranted) {
-      this.message.warning($localize`:@@permissions.no-permissions-to-visit-iam-page:You don't have permissions to visit IAM, please contact the admin to grant you the necessary permissions`);
-      return false;
-    }
-
-    return isGranted;
-  }
+  return isGranted;
```
deleteLater: `return true`

modules/front-end/src/app/core/guards/relayProxies.guard.ts
```diff
@@ -1,35 +1,21 @@
...
-    const isGranted = !!this.permissionsService.isGranted(generalResourceRNPattern.relayProxy, permissionActions.ListRelayProxies);
-
-    if (!isGranted) {
-      this.message.warning($localize`:@@permissions.no-permissions-to-visit-relay-proxy-page:You don't have permissions to view the relay proxy page, please contact the admin to grant you the necessary permissions`);
-      return false;
-    }
-
-    return isGranted;
-  }
+  return isGranted;
```
deleteLater: `return true`

PR #453, opened 2023-07-15
🐛 fix: organizations are not loaded after user refresh the page

cosmos-explorer:
For an authenticated user, if he refreshes the navigator, then go to the organization page, an error would happen in the console indicating that the curent org is undefined.


deleteLater (approved):
LGTM

PR #463, opened 2023-07-23
🏗 infra: UI log 404 instead of error when non-existing image, css and js files are requested

cosmos-explorer:
Currently when a non existing resource file (image, css, js) is requested, nginx would log a 404 along with an error. This behavior would trigger an alarm. The current PR force nginx only log the 404 and remove the error log


apo-j (approved):
LGTM

PR #465, opened 2023-07-24
🏗 infra: improve nginx settings

deleteLater:
Refs
- https://github.com/alibaba/Alink/blob/master/webui/tools/web/nginx.conf
- https://github.com/GoogleCloudPlatform/php-docker/blob/master/php-base/nginx.conf


cosmos-explorer (approved):
LGTM

PR #480, opened 2023-08-22
🐛 fix: update rule condition value didn't generate flag instructions

cosmos-explorer:
If we only update one condition's op or value, current code would not create any FlagInstructions, but a **UpdateRuleCondition** instruction should be created in this case.

Reproduce:
update rule condition: name is true => name is false


deleteLater (approved):
LGTM

PR #481, opened 2023-08-23
🐛 fix variation rollout compare bug

cosmos-explorer:
The FlagComparer would always return UpdateDefaultRuleVariationOrRolloutInstruction if both original and new values are percentage rollouts even they have the same value


deleteLater (approved):
LGTM

PR #482, opened 2023-08-23
🧹 refactor: implement a new version of segment/feature flag compare system

cosmos-explorer:
Pending changes
image

Review changes
image

Auditlog list
image


modules/back-end/src/Domain/SemanticPatch/FlagComparer.cs
```diff
@@ -147,13 +161,18 @@ public static IEnumerable<FlagInstruction> CompareFallthrough(Fallthrough origin
...
 
         if (isFallThroughChanged)
         {
-            instructions.Add(new DefaultVariationInstruction(current));
+            var rolloutVariations = new DefaultRuleRolloutVariations { RolloutVariations = current.Variations };
+            instructions.Add(new UpdateDefaultRuleVariationOrRolloutInstruction(rolloutVariations));
+        }
+
+        // compare dispatch key
+        if (original.DispatchKey != current.DispatchKey && current.DispatchKey != null)
```
deleteLater, line 173: What if we unset the `DispatchKey`, in this case `current.DispatchKey` will be null ?

modules/back-end/src/Domain/SemanticPatch/FlagComparer.cs
```diff
@@ -306,13 +326,6 @@ public static IEnumerable<FlagInstruction> CompareRule(TargetRule original, Targ
             instructions.Add(new RuleNameInstruction(value));
         }
 
-        // compare dispatch key
-        if (original.DispatchKey != current.DispatchKey)
-        {
-            var value = new RuleDispatchKey { RuleId = ruleId, DispatchKey = current.DispatchKey };
-            instructions.Add(new RuleDispatchKeyInstruction(value));
-        }
-
```
deleteLater, line 315: Any reason to move the dispatch key comparison down?

modules/back-end/src/Domain/SemanticPatch/Instruction.Flag.Rule.cs
```diff
@@ -98,4 +98,21 @@ public override void Apply(FeatureFlag flag)
...
+    }
+
+    public override void Apply(FeatureFlag flag)
+    {
+        if (Value is not string value)
+        {
+            return;
+        }
+
+        var rule = flag.Fallthrough.DispatchKey = value;
```
deleteLater: `var rule` is redundant

modules/back-end/src/Domain/SemanticPatch/FlagInstructionKind.cs
```diff
@@ -28,8 +28,10 @@ public static class FlagInstructionKind
 
     public const string UpdateDisabledVariation = nameof(UpdateDisabledVariation);
 
-    public const string UpdateDefaultVariation = nameof(UpdateDefaultVariation);
+    public const string UpdateDefaultRuleVariationOrRollouts = nameof(UpdateDefaultRuleVariationOrRollouts);
```
deleteLater, line 31: Any reason to change this name ?

modules/back-end/src/Domain/SemanticPatch/Instruction.Models.cs
```diff
@@ -41,6 +41,11 @@ public class RuleConditionValues : TheRuleId
...
 
 public class RuleVariations : TheRuleId
+{
+    public ICollection<RolloutVariation> RolloutVariations { get; set; }
+}
+
+public class DefaultRuleRolloutVariations
 {
     public ICollection<RolloutVariation> RolloutVariations { get; set; }
 }
```
deleteLater, line 51: Is it necessary to create a `DefaultRuleRolloutVariations` class ?

modules/back-end/src/Domain/SemanticPatch/Instruction.Flag.Variation.cs
```diff
@@ -76,17 +76,17 @@ public override void Apply(FeatureFlag flag)
     }
 }
 
-public class DefaultVariationInstruction : FlagInstruction
+public class UpdateDefaultRuleVariationOrRolloutInstruction : FlagInstruction
 {
-    public DefaultVariationInstruction(Fallthrough value) : base(FlagInstructionKind.UpdateDefaultVariation, value)
+    public UpdateDefaultRuleVariationOrRolloutInstruction(DefaultRuleRolloutVariations value) : base(FlagInstructionKind.UpdateDefaultRuleVariationOrRollouts, value)
```
deleteLater, line 81: I would prefer use `DefaultVariationsInstruction`

modules/front-end/src/app/core/components/change-review/change-review.component.ts
```diff
@@ -19,14 +19,13 @@ export class ChangeReviewComponent implements OnChanges {
   @Input() previous: string = '{}';
   @Input() current: string = '{}';
   @Input() refType: RefTypeEnum;
-  @Input() refs: IRefType;
   @Output() onSave = new EventEmitter<any>();
   @Output() onCancel = new EventEmitter<any>();
 
   title: string;
   reviewModalKindEnum = ReviewModalKindEnum;
   numChanges = 0;
```
deleteLater: Unused parameter `numChanges`

modules/front-end/src/app/core/components/change-list/instructions/add-rule-conditions/add-rule-conditions.component.ts
```diff
@@ -0,0 +1,84 @@
...
+
+    this.segmentRefs = {
+      ...this.segmentRefs,
+      ...segments.reduce((acc, cur) => {
+        acc[cur.id] = cur;
+        return acc;
+      }, {})
+    };
+  }
+}
```
deleteLater: Can we extract the `getSegmentRefs` method into `change-list\instructions\utils.ts`, then

```typescript
import { ISegment } from "@features/safe/segments/types/segments-index";
import { lastValueFrom } from "rxjs";
import { inject } from "@angular/core";
import { SegmentService } from "@services/segment.service";

export async function getSegmentRefs(segmentIds: string[]): Promise<{ [key: string]: ISegment }> {
  if (segmentIds.length === 0) {
    return {};
  }

  const segments = await lastValueFrom(inject(SegmentService).getByIds(segmentIds));
  return {
    ...segments.reduce((acc, cur) => {
      acc[cur.id] = cur;
      return acc;
    }, {})
  };
}
```

deleteLater: Since it's duplicate manywhere

image

modules/back-end/src/Domain/SemanticPatch/FlagComparer.cs
cosmos-explorer: If you set a rule from percentage rollout to a specific variation, yes, the DispatchKey would be set to null

modules/back-end/src/Domain/SemanticPatch/FlagComparer.cs
cosmos-explorer: We usually say: "serve true: 70%, false: 30% Dispatch by xxx"
image

modules/back-end/src/Domain/SemanticPatch/FlagInstructionKind.cs
cosmos-explorer: This is to keep the name consistant with others.
ex: UpdateRuleVariationOrRollouts
In addition, DefaultVariation is not splitted to **DefaultRuleVariationOrRollout** and **DefaultRuleDispatchKey**

modules/back-end/src/Domain/SemanticPatch/Instruction.Models.cs
cosmos-explorer: There is no ruleId here, so I don't want to reuse RuleVariations class

modules/back-end/src/Domain/SemanticPatch/Instruction.Flag.Variation.cs
cosmos-explorer: DefaultVariation is not splitted to **DefaultRuleVariationOrRollout** and **DefaultRuleDispatchKey**

modules/front-end/src/app/core/components/change-review/change-review.component.ts
cosmos-explorer: ok

modules/back-end/src/Domain/SemanticPatch/FlagComparer.cs
```diff
@@ -351,6 +363,13 @@ public static IEnumerable<FlagInstruction> CompareRule(TargetRule original, Targ
             instructions.Add(new UpdateVariationOrRolloutInstruction(variations));
         }
 
+        // compare dispatch key
+        if (original.DispatchKey != current.DispatchKey)
```
deleteLater: Maybe we need to add the `current.DispatchKey != null` check here ?

deleteLater:
We've introduced many duplicated CSS changes in this PR, we may need to fix later.

modules/front-end/src/app/core/components/change-review/change-review.component.ts
```diff
@@ -79,7 +80,9 @@ export class ChangeReviewComponent implements OnChanges {
         }
       });
 
-      return;
+      if (this.form.invalid) {
+        return;
+      }
```
deleteLater, line 84: Why this change?

deleteLater (approved):
LGTM

PR #487, opened 2023-09-04
🐛 fix: auditlog parse json error

cosmos-explorer:
Previously, when creating or deleting a feature flag, we don't generate any instructions, but the change-list component was still initiated which causes an error. 
image

The current fix does not instantiate the component if no instructions generated.


deleteLater (approved):
LGTM

PR #488, opened 2023-09-05
🐛 fixed redundant items in the pending scheduled changes drawer

cosmos-explorer:
Reproduce:
1. Create a schedule for a flag
2.  Click the icon 
image
3. One scheduled change should be visible
4. Create another schedule on the same flag (DO NOT REFRESH THE PAGE)
5. View the pending changes and you should be able to see three pending changes, among which one is redundant.


cosmos-explorer:
fixed in https://github.com/featbit/featbit/pull/489

PR #490, opened 2023-09-06
🐛 fix: organization undefined bug when the database is restored

cosmos-explorer:
Reproduce:
1. Log into an account
2. Restore the database
3. Refresh the browser
a blank page would be displayed and we could see the error in the console
image


deleteLater (approved):
LGTM

PR #500, opened 2023-10-02
🧹 refactor: enhance clickhouse cluster support

dsun0720:
1. fix bug for the migration in multiple ClickHouse nodes
2. support alternative hosts when getting results from ClickHouse
3. support round robin algorithm to get a ClickHouse node
4. support kafka multiple nodes cluster
5. support kafka user password


deleteLater (approved):
LGTM

PR #504, opened 2023-10-10
✨ feat: forbid sso users to reset password

cosmos-explorer:
SSO users comes from external idp, we don't store their password locally thus they should not be able to change their password with FeatBit. If it is really needed, the action should be done on the IDP side


deleteLater (approved):
LGTM

PR #508, opened 2023-10-14
✨ feat: disable organization creation

cosmos-explorer:
The present PR disabled the org creation to make it compatible with SSO and other features. Currently we put it under the enterprise license's protection. We may reenable this feature when we get a better vision. For users who really need this feature, please contact us to get license.


deleteLater (approved):
LGTM

PR #516, opened 2023-10-20
✨ feat: fallback to mongodb when redis is unavailable

deleteLater:
Currently, our evaluation server depends on Redis store only, this PR allows us to fallback to use MongoDB store when Redis store is unavailable


deleteLater:
We may add an in-memory cache later

cosmos-explorer (approved):
LGTM

PR #519, opened 2023-10-20
🧹 refactor: replace AnyAsync with CountAsync

### LINKED ISSUE #513: [Bug]: ".Any()" Operation aren't compatible with Azure CosmosDb
### What happened?

Hello,

We are running into an issue with creating FeateFlags, we are using Azure Cosmos DB for MongoDB (version 4.2)
When we initially create some feature flags, everything works as expected. If we create a new Organasation, everything works as expected. After a while though, trying to create a feature flag fails on validation the key. The UI reports Internal Server Error.

Debugging the C# code I Get:
`MongoDB.Driver.MongoCommandException: 'Command aggregate failed: [ActivityId=d1ac3449-de90-4e3d-aa05-07f096729d93] Internal error..'`

In the "FeatureFlagServic.HasKeyBeenUsedAsync(Guid envId, string key)" method on the "Any" call.

[Looking into this](https://github.com/Azure/azure-cosmos-dotnet-v3/issues/899), it looks like CosmosDB does not support the "Any" method when providing an expression (just "Any()" works).
The [official documentation](https://learn.microsoft.com/en-us/azure/cosmos-db/mongodb/how-to-dotnet-get-started?tabs=azure-cli%2Cwindows) seems to suggest using the mongodb.driver package that is already used. Updating to the latest version of this package does not fix this issue.
	
I tried locally with using Count == 0 or FirstOrDefault == null, either works.

### To Reproduce

image

image

### Expected behavior

_No response_

### Additional context

_No response_

### Code of Conduct

deleteLater:
For improved compatibility with CosmosDB, use CountAsync instead of AnyAsync. Thanks @Erik247ts for pointing this out.

fixed #513

### DIFF
--- a/modules/back-end/src/Infrastructure/Bases/MongoDbService.cs
+++ b/modules/back-end/src/Infrastructure/Bases/MongoDbService.cs
@@ -44,7 +44,8 @@ public class MongoDbService<TEntity> : IService<TEntity> where TEntity : Entity
 
     public async Task<bool> AnyAsync(Expression<Func<TEntity, bool>> predicate)
     {
-        return await Queryable.AnyAsync(predicate);
+        // for improved compatibility with CosmosDB, use CountAsync instead of AnyAsync.
+        return await Queryable.CountAsync(predicate) > 0;
     }
 
     public async Task AddOneAsync(TEntity entity)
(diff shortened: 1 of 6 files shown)


cosmos-explorer (approved):
LGTM

PR #528, opened 2023-10-28
✨ feat: support asterisk in license feature list

cosmos-explorer:
Previously, we need to list all features in the license. User need to regenerate the license whenever we add a new feature.
To avoid this, the current PR supports **asterisk(*)** for all features.


deleteLater (approved):
LGTM

PR #529, opened 2023-10-31
🧹 refactor: change the way audit logs are written

deleteLater:
This new way helps us eliminate more code and makes it easier to maintain. In addition to this, this PR also deletes the segment cache when a segment is deleted.


cosmos-explorer (approved):
LGTM

PR #530, opened 2023-11-02
🧹 refactor: rename account to organization

cosmos-explorer:
The present PR has two purposes:

1.  The account here is previously indeed organization, so it's better to rename it to organization
2. We are going to introduce a new entity: **Account**, it would be a higher level of organization. This would be done in a future PR very soon

This PR brings some breaking changes, be sure to run the following code against your mongodb:
```javascript
const cursor = db.Policies.find({ "statements.resourceType": "account" });

cursor.forEach(doc => {
  doc.statements.forEach(statement => {
    if (statement.resourceType === "account") {
      statement.resourceType = "organization";
      statement.resources = statement.resources.map(resource => resource.replace(/^account\//, "organization/"));
    }
  });

  db.Policies.updateOne({ _id: doc._id }, { $set: { statements: doc.statements } });
});
```

### DIFF
--- a/infra/mongodb/docker-entrypoint-initdb.d/init.js
+++ b/infra/mongodb/docker-entrypoint-initdb.d/init.js
@@ -97,10 +97,10 @@ db.Policies.insertOne(
         statements: [
             {
                 _id: getUUIDString(),
-                resourceType: "account",
+                resourceType: "organization",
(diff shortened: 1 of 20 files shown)


deleteLater (approved):
LGTM

PR #532, opened 2023-11-07
🖍 style: extend length of selection item content of nz-select component

### LINKED ISSUE #506: [Bug][Good First Issues]: Inconvinient to overview my feature flag configuration
### What happened?

We have a lot of user segments, we have used it to do some beta testing, experimental features and also AB testing.
I am trying to use FeatBit to solve the orthogonality requirement of our AB testing.

But we all find it very inconvenient that we can't easily see our segment's name.  As the figure below shows, I can't easily see the difference between each segment.

I'd say that the length of the segment name can easily exceed the display limit you set by default.

I hope you can solve this little problem in the near future.

:D

### To Reproduce

image

### Expected behavior

_No response_

### Additional context

_No response_

### Code of Conduct

mucoban:
Length of selection item content of nz-select component are extended so segment titles wouldn't exceed its limit easily
Screenshot 2023-11-07 at 13 44 05

fixed #506

### DIFF
--- a/modules/front-end/src/locale/messages.xlf
+++ b/modules/front-end/src/locale/messages.xlf
@@ -2634,6 +2634,88 @@
           <context context-type="linenumber">34</context>
         </context-group>
       </trans-unit>
+      <trans-unit id="core.components.findrule.if" datatype="html">
+        <source>If</source>
+        <context-group purpose="location">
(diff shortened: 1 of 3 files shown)


cosmos-explorer (approved):
LGTM

PR #533, opened 2023-11-08
✨ feat: webhooks

deleteLater:
Allow users to subscribe to flag and segment changes


deleteLater:
There will have some other webhook features be done in other PRs

- add human-readable changes description to our webhook payload
- view webhook logs

modules/front-end/src/app/core/services/handlebars.service.ts
```diff
@@ -0,0 +1,32 @@
...
+@Injectable({
+  providedIn: 'root',
+})
+
+export class HandlebarsService {
+
+  private _initialized = false;
+
+  init() {
+    if (this._initialized) {
```
cosmos-explorer, line 14: Can't we put this logic in constructor? Services are singletons

modules/front-end/src/app/core/services/monaco-service.ts
```diff
@@ -0,0 +1,65 @@
...
+@Injectable({
+  providedIn: 'root',
+})
+
+export class MonacoService {
+
+  private _initialized = false;
+
+  init() {
+    if (this._initialized) {
```
cosmos-explorer, line 15: Can't we put this logic in constructor? Services are singletons

cosmos-explorer:
image
Maybe use another separator here?

cosmos-explorer (approved):
LGTM

PR #539, opened 2023-11-10
🖍 style: fix select component style in flag details page

deleteLater:
This PR

- fixed the select component's selection item not centered problem
- reverting an abnormal i18n change introduced by #532

## Before

before
before-2

## After

after
after-2


cosmos-explorer (approved):
LGTM

Issue #370, opened 2023-05-18
[Good First Issues]: Fix CommonJS dependency warnings


deleteLater:
# Description

If you run our frontend project with `npm run start`, you can see these warnings. We're planning to fix these warnings.

![image](https://github.com/featbit/featbit/assets/34052208/3d75e63d-2ec3-4f87-b8bc-350302d7e977)

# Scope

- Service: UI

# Setup dev environment

Please read the [doc](../blob/main/Development.md) to set up your development environment.

# Deadline

We usually give 2 weeks for a feature request, which means you have 2 weeks to implement the feature starting from the
day it is assigned to you. If you have any difficulties, please contact us in [Slack](https://join.slack.com/t/featbit/shared_invite/zt-1ew5e2vbb-x6Apan1xZOaYMnFzqZkGNQ).

# How to implement

1. Fork the FeatBit repository.
2. Create a branch issues-{issues number}
3. Implement the task based on Description and Scope mentioned above.
4. Submit your PR

If the feature is UI related, make sure you complete the following tasks before submit your PR:

- merge featbit main branch into your working branch
- run `npm run i18n`
- commit **messages.xlf & messages.zh.xlf** file changes (if any)

# Contribution points

Each contribution counts a certain number of points, depending on its difficulty. Contributors earn the appropriate
points when the work is merged. Contribution points are used to describe the contributions that contributors have made.
They can also be used to receive rewards for community events. You can view the current community submissions on a
public [google sheet](https://docs.google.com/spreadsheets/d/1ukyXgi_jRPeXj7EAST0IrnPfLOQ6xDBkcyAJY9N-Yb4/edit#gid=1117970540).

Points: 12

# How to claim to solve the issue

If you want to implement this function, please leave a comment in this issue like:

> I'd like to implement this function, please assign this issue to me.

cosmos-explorer:
@defqd just assigned the issue to you, thank you

deleteLater:
Hi @defqd :wave:

I recommend that you separate the solution into multiple PRs, one for each of the following:

- lodash
- momentjs
- prismjs

This will make it easier for us to focus on each change individually, and it will also make it easier to merge the changes once they have been reviewed.

deleteLater:
Replace `loadsh` with `lodash-es`: #378

yashchaddha:
I would like to work on this issue, hence would like to request for it to be assigned to me 
Thanks in advance

deleteLater:
[...]

PR #540, opened 2023-11-12
🧹refactor(front-end): add allowedCommonJsDependencies to angular.json

fMoro1999:
The following PR aims to remove the annoying allowed common js dependencies warning (#370)

### DIFF
--- a/modules/front-end/angular.json
+++ b/modules/front-end/angular.json
@@ -55,7 +55,11 @@
                 "src/"
               ]
             },
-            "scripts": []
+            "scripts": [],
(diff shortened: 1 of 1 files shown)


cosmos-explorer:
Hello @fMoro1999 , what we planned to do for this issue is to fix the warning instead of ignoring it. You can for example try to find an alternative of momentjs, date-fns (https://date-fns.org/) for example, and replace all usage of momentjs

fMoro1999:
My bad, I misunderstood. Maybe, when I have more time I will give it a go! Thank you by the way!

PR #542, opened 2023-11-13
🐛 fix: trigger didn't update flag correctly

deleteLater:
We need to update the `updatedAt` field when trigger runs, otherwise our SDK will not accept the flag change since they have the same update timestamp.


cosmos-explorer (approved):
LGTM

PR #543, opened 2023-11-13
🖍 style: unify nz-select style

cosmos-explorer:
The styles for nz-select was a messy. Any changes to one may break others. Some of the text/selected item(s)/placeholder are not vertically centered, it was a cauchmare to fix it. So we decided to refactor the styles for all nz-select.


deleteLater (approved):
LGTM

PR #549, opened 2023-11-29
🏗 infra: refactor publish-docker-images.yml

deleteLater:
This PR

- Always run publish image GitHub Action manually since we need to decide whether to publish latest image
- Make publish-docker-image.yml simpler
- Improving the workflow speed; previously taking 45 minutes, the process now completes in just 22 minutes.

Refer [this](https://github.com/featbit/featbit/actions/runs/7043363396) to check out an example run. Additionally, compare the performance with a previous run (v3.0.0) [here](https://github.com/featbit/featbit/actions/runs/6906276351).


cosmos-explorer (approved):
LGTM

PR #559, opened 2023-12-21
✨ feat: add instruction desctiptor

deleteLater:
Use instruction descriptor to describe instruction change

PR #563, opened 2023-12-26
✨ feat: add webhook change description

deleteLater:
add webhook change description to webhook callback, follow up #533 

based on #559


cosmos-explorer:
This logic is missing when live debug a webhook

cosmos-explorer (approved):
LGTM

PR #583, opened 2024-01-26
🐛 Bugfix: When operation is selected, property cannot be selected.

### LINKED ISSUE #576: [Bug]: When select operation first, the property can't be selected
### What happened?

图片

### To Reproduce

1. Go to feature flag
2. Add a targeting rules
3. Select operation as "is one of"
4. Select property will have empty content

### Expected behavior

Maybe disable the "select operation" at first or make sure the content in the "select property" is correct.

### Additional context

_No response_

### Code of Conduct

qixiaojian310:
fixed #576

Current effect: 
图片
图片
图片


cosmos-explorer (approved):
LGTM

PR #587, opened 2024-02-05
Add mongo details to docker pro for da-server. 🏗

adambirds:
Add mongo details to docker pro for da-server as they were missing.

### DIFF
--- a/docker-compose-pro.yml
+++ b/docker-compose-pro.yml
@@ -65,6 +65,9 @@ services:
       KAFKA_HOSTS: kafka:9092
       REDIS_HOST: redis
       REDIS_PORT: 6379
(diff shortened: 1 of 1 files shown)


cosmic-flood:
Thanks for the PR. We're investigating why we need MongoDB variables for da-server, because it was supposed that da-server doesn't communicate with MongoDB for pro version. Working in progress...

deleteLater:
fixed in https://github.com/featbit/featbit/pull/595/files#diff-eef525458fcc0bff240b53a2df99c87116de574cae540983e87f5867d61bacca

deleteLater:
This bug has been fixed in v3.2.1

PR #595, opened 2024-02-19
🐛 fix: da-server otel integration

dsun0720:
- optimize otel auto Instrumentation configure in da server
- fix the bug of migration in pro mode
- fix the bug: api and evaluation server can't start up in arm64 arch when enable otel


modules/data-analytics/start.sh
deleteLater, line 1: Prefer use `=` instead of `==`. Check https://stackoverflow.com/questions/20449543/shell-equality-operators-eq

modules/back-end/deploy/Dockerfile
deleteLater, line 1: Why this change?

modules/data-analytics/.dockerignore
deleteLater, line 1: Seems to no reason for this change

modules/data-analytics/app/health/views.py
deleteLater, line 1: unused change

modules/data-analytics/.dockerignore
dsun0720: integration_tests is just used to test in local, not fit in docker image

modules/back-end/deploy/Dockerfile
dsun0720: api and els should be able to start in arm64. https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/issues/2181

modules/data-analytics/app/health/views.py
dsun0720: flake 8 format

deleteLater (approved):
LGTM

PR #605, opened 2024-03-07
✨ Add ClientHost, ClientIP, EnvId to eval server logs

### LINKED ISSUE #604: [Good First Issues]: Add Connection Information to Featbit Eval Server logs
# Description
Currently, the logging functions defined in `ConnectionHandler.Log.cs` class only log the `connectionId`, and error message if there was one, and in some cases an event.

`connectionId` is not very helpful in tracking down which client interaction caused the log message.

The logs should include the ~~project id~~, the environment id, client IP address, and client hostname if availabe.

With this information system admins will be able to find the project, environment, and client associated with a log message

The client IP address can be obtained in 'StreamingMiddleware.cs' by capturing one of the following (see references)

`context.Connection.RemoteIpAddress`
`HttpContext.GetServerVariable("HTTP_X_FORWARDED_FOR");`
`Request.Headers["CF-CONNECTING-IP"];`

The hostname can be retrieved with
`Dns.GetHostEntry(<ip address variable>).HostName;`

These should be passed to `RequestHandler` and added to the `Connection` Object, Connections should not fail or throw exceptions if the client IP, or hostname are not retrieveable, they should instead be empty strings.

~~The projectId can be obtained using the MongoDbClient with a query similar to the following~~

~~The projectId retrieval ideally will be feature gated using an environment variable such as `EVAL_LOG_PROJECT_ID` with a default of false. this is to avoid adding potential latency to the initial websocket connection while this feature matures.~~

`LoggerMessage` Attriubutes and associated Funcitons in `ConnectionHandler.Log.cs` should be updated to include `envId`, ~~`projectId`~~, `clientId`, and `clientHostname`

Refactor function callers to include this information.

#References

https://juldhais.net/how-to-get-client-ip-address-and-location-information-in-asp-net-core-c2bb50e689c3

# Scope

Evaluation Server

# Screenshot

# Setup dev environment

Please read the [doc](../blob/main/Development.md) to set up your development environment.

# Deadline

We usually give 2 weeks for a feature request, which means you have 2 weeks to implement the feature starting from the
day it is assigned to you. If you have any difficulties, please contact us in [Slack](https://join.slack.com/t/featbit/shared_invite/zt-1ew5e2vbb-x6Apan1xZOaYMnFzqZkGNQ).

# How to implement

1. Fork the FeatBit repository.
2. Create a branch issues-{issues number}
3. Implement the task based on Description and Scope mentioned above.
4. Submit your PR

If the feature is UI related, make sure you complete the following tasks before submit your PR:

- merge featbit main branch into your working branch
- run `npm run i18n`
- commit **messages.xlf & messages.zh.xlf** file changes (if any)

# Contribution points

Each contribution counts a certain number of points, depending on its difficulty. Contributors earn the appropriate
points when the work is merged. Contribution points are used to describe the contributions that contributors have made.
They can also be used to receive rewards for community events. You can view the current community submissions on a
public [google sheet](https://docs.google.com/spreadsheets/d/1ukyXgi_jRPeXj7EAST0IrnPfLOQ6xDBkcyAJY9N-Yb4/edit#gid=1117970540).

Points: 4

# How to claim to solve the issue

If you want to implement this function, please leave a comment in this issue like:

> I'd like to implement this function, please assign this issue to me.

wss-alubic:
ClientIP, ClientHost, and EnvId are logged in the ConnectionHandler.  Tests were added for ConnectionHandler to ensure functionality works as expected.

EnvId was chosen over EnvKey because the UI does not allow searching on env.

Once work is done to add ProjectId/ProjectKey to Connection, we could add project related info to the logs in a future PR.

fixes #604

### DIFF
--- a/modules/evaluation-server/src/Streaming/StreamingMiddleware.cs
+++ b/modules/evaluation-server/src/Streaming/StreamingMiddleware.cs
@@ -1,4 +1,6 @@
-﻿using System.Net.WebSockets;
+﻿using System.Net;
+using System.Net.WebSockets;
+using Infrastructure.MongoDb;
 using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Internal;
@@ -42,9 +44,11 @@ public class StreamingMiddleware
         // try accept request
         var query = request.Query;
         var currentTimestamp = _systemClock.UtcNow.ToUnixTimeMilliseconds();
+        var clientIpAddress = GetClientIpAddress(context);
+        var clientHost = await GetClientHost(clientIpAddress);
 
         var connection =
-            RequestHandler.TryAcceptRequest(ws, query["type"], query["version"], query["token"], currentTimestamp);
+            RequestHandler.TryAcceptRequest(ws, query["type"], query["version"], query["token"], currentTimestamp, clientIpAddress, clientHost);
         if (connection == null)
         {
             await ws.CloseOutputAsync(
@@ -57,5 +61,47 @@ public class StreamingMiddleware
 
         // use ApplicationStopping token
         await handler.OnConnectedAsync(connection, _applicationLifetime.ApplicationStopping);
+    }
+
+    private string GetClientIpAddress(HttpContext context)
+    {
+        if (context.Request.Headers.TryGetValue("X-Forwarded-For", out var forwardForHeaders))
+        {
+            return forwardForHeaders.First();
+        }
+        // cloudflare connecting IP header
+        // https://developers.cloudflare.com/fundamentals/reference/http-request-headers/#cf-connecting-ip
+        if (context.Request.Headers.TryGetValue("CF-Connecting-IP", out var cfConnectingIpHeaders))
+        {
+            return cfConnectingIpHeaders.First();
+        }
+        
+        var remoteIpAddr = context.Connection.RemoteIpAddress?.ToString();
+        if (!string.IsNullOrWhiteSpace(remoteIpAddr))
+        {
+            return remoteIpAddr;
+        }
+
+        return "";
+    }
+
+    private async Task<string> GetClientHost(string clientIpAddress)
+    {
+        var clientHost = "";
+        if (string.IsNullOrEmpty(clientIpAddress))
+        {
+            return clientHost;
+        }
+        
+        try
+        {
+            clientHost = (await Dns.GetHostEntryAsync(clientIpAddress)).HostName;
+        }
+        catch (Exception)
+        {
+            // allow clientHost to stay empty without failing the connection.
+            clientHost = "";
+        }
+        return clientHost;
     }
 }
--- a/modules/evaluation-server/src/Streaming/Connections/Connection.cs
+++ b/modules/evaluation-server/src/Streaming/Connections/Connection.cs
@@ -19,6 +19,10 @@ public class Connection
     /// </summary>
     public EndUser? User { get; set; }
 
+    public string ClientIpAddress { get; }
+    
+    public string ClientHost { get; }
+    
     public string Type { get; }
 
     public string Version { get; }
@@ -33,7 +37,9 @@ public class Connection
         string type,
         string version,
         long? connectAt = null,
-        long? closeAt = null)
+        long? closeAt = null,
+        string clientIpAddress = "",
+        string clientHost = "")
     {
         Id = Guid.NewGuid().ToString("D");
 
@@ -43,6 +49,8 @@ public class Connection
         Version = version;
         ConnectAt = connectAt ?? DateTimeOffset.UtcNow.ToUnixTimeMilliseconds();
         CloseAt = closeAt ?? 0;
+        ClientIpAddress = clientIpAddress;
+        ClientHost = clientHost;
     }
 
     public async Task SendAsync(Message message, CancellationToken cancellationToken)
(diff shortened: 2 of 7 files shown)


modules/evaluation-server/src/Streaming/StreamingMiddleware.cs
```diff
@@ -1,4 +1,6 @@
-﻿using System.Net.WebSockets;
+﻿using System.Net;
+using System.Net.WebSockets;
+using Infrastructure.MongoDb;
```
deleteLater: remove unused using `using Infrastructure.MongoDb`

modules/evaluation-server/src/Streaming/StreamingMiddleware.cs
```diff
@@ -58,4 +62,46 @@ public async Task InvokeAsync(HttpContext context, IConnectionHandler handler)
...
+            return cfConnectingIpHeaders.First();
+        }
+        
+        var remoteIpAddr = context.Connection.RemoteIpAddress?.ToString();
+        if (!string.IsNullOrWhiteSpace(remoteIpAddr))
+        {
+            return remoteIpAddr;
+        }
+
+        return "";
```
deleteLater: How about

```csharp
var remoteIpAddr = context.Connection.RemoteIpAddress?.ToString();
return remoteIpAddr ?? string.Empty;
```

modules/evaluation-server/src/Streaming/StreamingMiddleware.cs
```diff
@@ -58,4 +62,46 @@ public async Task InvokeAsync(HttpContext context, IConnectionHandler handler)
...
+        if (context.Request.Headers.TryGetValue("X-Forwarded-For", out var forwardForHeaders))
+        {
+            return forwardForHeaders.First();
+        }
+        // cloudflare connecting IP header
+        // https://developers.cloudflare.com/fundamentals/reference/http-request-headers/#cf-connecting-ip
+        if (context.Request.Headers.TryGetValue("CF-Connecting-IP", out var cfConnectingIpHeaders))
+        {
+            return cfConnectingIpHeaders.First();
+        }
```
deleteLater: Compared to `First()`, I prefer `FirstOrDefault(string.Empty)`

```
return headers.FirstOrDefault(string.Empty);
```

modules/evaluation-server/src/Streaming/StreamingMiddleware.cs
```diff
@@ -58,4 +62,46 @@ public async Task InvokeAsync(HttpContext context, IConnectionHandler handler)
...
+        {
+            clientHost = (await Dns.GetHostEntryAsync(clientIpAddress)).HostName;
+        }
+        catch (Exception)
+        {
+            // allow clientHost to stay empty without failing the connection.
+            clientHost = "";
+        }
+        return clientHost;
+    }
```
deleteLater: We can make it a little bit cleaner.

```csharp
private async Task<string> GetClientHostAsync(string clientIpAddress)
{
    if (string.IsNullOrEmpty(clientIpAddress))
    {
        return string.Empty;
    }

    try
    {
        return (await Dns.GetHostEntryAsync(clientIpAddress)).HostName;
    }
    catch (Exception)
    {
        return string.Empty;
    }
}
```

Another thing I was wondering is if we need to add a timeout to the `Dns.GetHostEntryAsync` method to avoid  DNS temporary failure? something like: 

```csharp
// 50ms for example
var cancellationToken = new CancellationTokenSource(TimeSpan.FromMilliseconds(50)).Token;
return (await Dns.GetHostEntryAsync(clientIpAddress, cancellationToken)).HostName;
```

deleteLater:
Thanks for your time on this PR ❤️ @wss-alubic . I've left some comments that we can disccuss with.

modules/evaluation-server/src/Streaming/StreamingMiddleware.cs
wss-alubic: I cleaned up the method and I agree that a timeout should be added, but 50ms seems too aggressive.

[IETF RFC 1123](https://www.rfc-editor.org/rfc/rfc1123#page-77) doesn't mandate specific timeout values, but recommends no less than 5 seconds.

nslookup defaults to 5 seconds, i believe.  I'm seeing some guidance across the web that 2 seconds might be reasonable.  How do you feel about 2000ms?

modules/evaluation-server/src/Streaming/StreamingMiddleware.cs
wss-alubic: Done

modules/evaluation-server/src/Streaming/StreamingMiddleware.cs
wss-alubic: Done.  Used FirstOrDefault with string.Empty as default.

modules/evaluation-server/src/Streaming/StreamingMiddleware.cs
wss-alubic: Done.  sorry about that.  This PR previously had logic to get the projectId before I submitted it, but I removed that in favor of your future PR that'll put more info on the connection secret instead.

modules/evaluation-server/src/Streaming/StreamingMiddleware.cs
deleteLater: Either 5 or 2 seconds is fine to me.

wss-alubic: Added cancellation token

deleteLater (approved):
LGTM, thank you @wss-alubic

PR #609, opened 2024-03-19
🧹 refactor: enhance secret check

deleteLater:
Check if the secret type is consistant with the connection type. That is, for server-side SDK connections, we must use the server secret; for client-side SDK connections, we must use the client secret.


cosmos-explorer (approved):
LGTM

PR #610, opened 2024-03-21
🚀 perf: add store availablility check sentinel

deleteLater:
Utilize `StoreAvailableSentinel` to avoid unnecessary store availability check

### DIFF
--- a/modules/evaluation-server/src/Infrastructure/Store/HybridStore.cs
+++ b/modules/evaluation-server/src/Infrastructure/Store/HybridStore.cs
@@ -11,6 +11,7 @@ public class HybridStore : IStore
 
     private readonly IStore _redis;
     private readonly IStore _mongodb;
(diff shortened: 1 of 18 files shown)


modules/evaluation-server/src/Infrastructure/Store/HybridStore.cs
```diff
@@ -1,101 +1,63 @@
...
-        foreach (var store in _stores)
+        IStore GetAvailableStore(string store)
         {
-            var isStoreAvailable = await store.IsAvailableAsync();
-            if (isStoreAvailable)
+            return store switch
             {
-                return await store.GetFlagsAsync(ids);
-            }
+                "Redis" => _redis,
```
cosmos-explorer: You can use `Stores.Redis`

cosmos-explorer (approved):
LGTM

PR #613, opened 2024-03-22
✨ feat: allow client SDKs to get incremental flag changes by API

cosmos-explorer:
For Client SDKs supporting polling, now we can get the incremental changes by adding a timestamp to the query string, example:
`https://xxx/api/public/sdk/client/latest-all?timestamp=1711096130599`


deleteLater (approved):
LGTM

PR #614, opened 2024-03-22
✨ feat: add `variationId` and `sendToExperiment` field to get client feature flags API

cosmos-explorer:
Client SDK needs `variationId` and `sendToExperiment` when sending insight messages


deleteLater:
We need to check our client-js-sdk to see if we also need to add `SendToExperiment` field to the response

deleteLater (dismissed):
LGTM

deleteLater (approved):

Issue #582, opened 2024-01-24
[BUG] If the variation name is not short, we won't be able to see the full name.


cosmic-flood:
The screenshot below describe the bug clearly.

- ChatGPT 4 32k
- ChatGPT 3.5 Turbo 16k
- OpenChat 3.5
- ...

![image](https://github.com/featbit/featbit/assets/68597908/6484675c-4df2-4e54-8466-15cc7848dc48)

deleteLater:
Adding a tooltip for the variation name would be useful. And for the copy and expand button, we just need one. For example, we left the expand button, and enabling users to copy the value by clicking it. @cosmos-explorer

qixiaojian310:
I'd like to implement this function, please assign this issue to me.

cosmos-explorer:
Fixed in https://github.com/featbit/featbit/pull/616

PR #616, opened 2024-03-25
🐛 Fix: If the variation name is not short, can't see the full name.

qixiaojian310:
Add a tooltip when hover the variation name, and remove copy button, when user click the tag they can copy variation value. 
1. Add a tooltip
图片
2. When click the tag, it will be copied
图片

(#582)


cosmos-explorer (approved):
LGTM

PR #618, opened 2024-03-27
🐛 fix: ui is blank if workspaceId is invalid

deleteLater:
This is not a common problem (only happens when switching databases), but let's just fix that for better development experience


cosmos-explorer (approved):
LGTM

PR #620, opened 2024-03-28
✨ feat: resource finder

deleteLater:
add a resource finder to make it easier to select resources (project, environment, flag, segment...)

example usage:

```html
<resource-finder
  [isVisible]="true"
  [resources]="[ResourceTypeEnum.Project, ResourceTypeEnum.Env]"
  [spaceLevel]="ResourceSpaceLevel.Organization"
  (onClose)="showSelectedResources($event)">
</resource-finder>
```


modules/back-end/src/Infrastructure/Resources/ResourceServiceV2.cs
```diff
@@ -0,0 +1,180 @@
...
+
+            resources.AddRange(items);
+        }
+
+        return resources;
+    }
+
+    private async Task<IEnumerable<ResourceV2>> GetOrganizationsAsync(Guid spaceId, string spaceLevel, string name)
+    {
+        if (spaceLevel == ResourceSpaceLevel.Organization)
```
cosmos-explorer: I think it would be better to use:
`if (spaceLevel != ResourceSpaceLevel.Workspace)`

deleteLater: resolved in https://github.com/featbit/featbit/pull/620/commits/c2defcc8a43d50999a691704617b501b8782bd25

cosmos-explorer (approved):
LGTM

PR #635, opened 2024-04-22
🧹 refactor: align get client feature flags API with streaming response format

cosmos-explorer:
return 

1. empty result when there is no data change
2. the following structure when there are changes for client side api
```
{
    "messageType": "data-sync",
    "data": {
        "eventType": "full",
        "userKeyId": "my-user",
        "featureFlags": []
    }
}
```


deleteLater (approved):
LGTM

PR #638, opened 2024-05-08
🐛 fix: fixed date formatting bug

cosmos-explorer:
Previously, when we create an experiment, the month was displayed as a random number, this is due to migration from momentjs to date-fns, and the format string was not correct. The current PR fixed this problem.


deleteLater (approved):
LGTM

PR #644, opened 2024-05-23
🐛fix: experiment data lose bug

cosmos-explorer:
Reproduce:

1.  Create and start a new experiment
2.  Stop the experiment
3.  Restart the experiment
4. Generate some data and wait until it shows in the table below
5. In the period selector, select the last period at the bottom of the dropdown
6. Stop the experiment
7. Change period to the first one in the list

You should see that data displayed in step 4 disappeared.

Other changes:

- Changed action button labels:
Restart => Start experiment
Pause => Stop experiment

- Remove the `(now)` string when a period is stopped
- Add radius to **Show chart** and **Hide chart** buttons


deleteLater (approved):
LGTM

PR #649, opened 2024-06-20
✨ feat: provide default workspace key for SSO

cosmos-explorer:
For self-hosted environnment, it's very annoying to provide the workspace key each time a user wants to login with SSO.
The current PR would fill this field by the default workspace, so user no need to worry about this.

image


deleteLater (approved):
LGTM

PR #651, opened 2024-07-23
✨ feat: fill variation values according to variation type

### LINKED ISSUE #650: Feature Suggestion: Filling variation values according to variation type in the create feature flag page
**Problem**
When creating a feature flag, change variation type from boolean to string, all variation values stay as 'true'. This can be confusing for users.
image

**Solution**
We can adjust the frontend as variation values be filled correctly upon the change of variation type in create feature flag page.
It can be like
When variation type is selected as 
boolean, 1st variation value is to be true and 2nd is to be false (it is like that now already)
image

string, all variation values are to be their variation names in lowercase
image

number,  all variation values are to be the numbers in their variation names
image

json,  all variation values are to be the {} or []
image

We can also update variation values while user typing in variation names,  according to above rules to help to user.

I can do this feature if I am assigned to it.

mucoban:
Fixed #650 by adjusting create feature flag page as variation value would be filled according to  variation name

### DIFF
--- a/modules/front-end/src/app/core/components/change-list/constants.ts
+++ b/modules/front-end/src/app/core/components/change-list/constants.ts
@@ -132,7 +132,7 @@ export const instructionCategories: ICategoryInstruction[] = [
   },
   {
     category: CategoryEnum.DefaultRule,
-    label: $localize`:@@common.default-rule:Default rule`,
+    label: $localize`:@@common.default-rule:Default Rule`,
     instructions: [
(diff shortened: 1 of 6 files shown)


cosmic-flood:
Hi @mucoban  we were in a big vacation. we will check it

mucoban:
hi @cosmic-flood, can you review the pull request? If you have additional requirements or adjuments I am willing to apply.

deleteLater:
Hi @mucoban , sorry for the late response. I've made some refactoring to your code. And I removed the `onInputVariationName` logic because I don't think that's necessary and the code looks too complicated to me.

deleteLater (approved):
LGTM

PR #673, opened 2024-10-16
🚀 perf: add mongodb indexes

MattJeanes:
Added MongoDB indexes to all relevant tables that are queried with a particular sorting key.

This improves performance, but is not actually the reason I made this change. This PR adds support for running on Azure CosmosDB for MongoDB.

Azure CosmosDB for MongoDB requires that indexes exist when querying using sorting keys, so before we added these to our setup, it would error when attempting to query the database.

Helm chart PR: https://github.com/featbit/featbit-charts/pull/40


cosmic-flood:
@deleteLater @apo-j

deleteLater:
LGTM, Thanks @MattJeanes !

deleteLater (approved):

PR #676, opened 2024-10-28
🐛 fix: switching environment bug

cosmos-explorer:
If user is on detail page of feature flag or segment or other object, when switching environment, user would stay on the detail page which is bug, the current PR reload the whole APP and navigate to home page.


deleteLater (approved):
LGTM

Issue #678, opened 2024-11-05
Feature Suggestion: Limit Webhooks to Specific Feature Flags


drewsteinacher:
**Is your feature request related to a problem? Please describe.**
My team uses a [Webhook](https://docs.featbit.co/integrations/webhooks) to send notifications to a Teams channel when only a _specific_ feature flag is toggled. Note that Webhooks fire for _all_ matching feature flag events and segments.

The only way I was able to implement this was by making clever use of Handlebars.js templating to only send an Adaptive Card body if `data.object.key` was the desired feature flag key, and an empty JSON body otherwise:
```
{
   {{#eq data.object.key "desired-key"}}
   // Desired Adaptive Card body goes here
   {{/eq}}
}
```

The empty JSON body is required, since this body must be valid JSON, or the Webhook form will not submit.
* [Frontend requirement](https://github.com/featbit/featbit/blob/065f9d040853fd66f16589509ef5793948acea17/modules/front-end/src/app/core/components/webhook-drawer/webhook-drawer.component.ts#L132-L138)
* [Backend requirement](https://github.com/featbit/featbit/blob/065f9d040853fd66f16589509ef5793948acea17/modules/back-end/src/Infrastructure/Webhooks/WebhookSender.cs#L40C13-L40C41)

When the empty JSON body is sent, the webhook returns a 400-level response (due to it not being a valid Adaptive Card), and no notification is sent, thus having the desired behavior.

However, due to the [impending MS Teams Connectors deprecation](https://devblogs.microsoft.com/microsoft365dev/retirement-of-office-365-connectors-within-microsoft-teams/), we have had to change how we send these notifications.
Our new in-house solution does not behave the same with the empty body and will send a "blank" notification in this case, and will otherwise send some kind of notification _no matter what body it receives_. This means that it would spam our Teams channel with a notification whenever _any_ feature flag was enabled/disabled. Due to internal requirements, this behavior _will not be changed_, meaning my previous templating-based solution will not work. 

There needs to be some way to limit Webhooks from firing for some feature flags on the FeatBit side that doesn't involve creating extra notification-only environments (which would cause clutter and confusion and has been deemed as an unacceptable solution by my team).

**Describe the solution you'd like**
I can think of a few that would be acceptable, listed in no particular order:
1. **Add an optional "Feature Flags" chips/multi-select input under the "Scopes" section of the Webhook editing pane**
- This naturally extends the existing "Project" and "Environments" that are already there.
[...]

PR #681, opened 2024-11-14
✨feat: optionally prevent empty webhook payloads

drewsteinacher:
## Summary
Implemented optional prevention of empty webhook payloads as described in #678.

It seems that I cannot add the labels: UI, API

## Testing instructions:
- Create two feature flags in an environment/project scope
	- e.g. keys like `feature-flag-1` and `feature-flag-2`
- Create a webhook
	- Use a valid URL for an API you can verify is being hit or not
	- Use the same project/environment scope as the feature flags
	- Select only the "Toggled" Feature Flag event (though other events should also work)
	- Use a custom payload template that gives an non-empty request for one of the feature flag keys, e.g.:
```
{
    {{#eq data.object.key "feature-flag-1"}}
    "featureFlagKey": "{{ data.object.key }}"
    {{/eq}}
}
```
- Check the "Prevent empty payloads from being sent" checkbox
- Ensure that the Active checkbox is selected
- Save
- Test via Live Debug
	- Open the webhook's Live Debug feature
	- Select an appropriate event (e.g. feature flag toggle)
	- Click "Send test webhook"
	- Confirm that the webhook API did not receive any payload
	- An error response should be presented
	- The Response tab should indicate that the payload was not sent due to settings
	- This could be repeated if the payload template is edited to allow a non-empty response for the test payload's feature flag key of `test`, looking for successful behavior instead of failures
- Test via Feature Flag Toggle
	- Toggle the feature flag that will give an empty response per the template
	- Confirm that the webhook API did not receive a payload
	- Open the webhook's logs (More > View Logs)
	- Confirm that an error response is recorded
	- The Response tab should indicate that the payload was not sent due to settings
	- Repeat for the other feature flag that will give a non-empty response per the template
	- Expect it to give a successful log and API payload should be received

#### Before
image

#### After (Checkbox added, defaults to unchecked)
image

#### Before (all requests are always sent)
image

image

#### After (empty responses can be prevented)
image

#### After (Prevented payloads are recorded)
image

## Other
See further discussion in #678.

<!--
# Instructions

1. Pick a meaningful and result oriented title for your pull request. (Use sentence case, don't include any words indicating the way you achived this result, we just want the result. Keep it short < 70 characters)
  - Prefix the title with an emoji to categorize what is being done. (Copy-paste the emoji from the list below.)
  - Assign the appropriate label(s) to the pull request. (Use one of the labels from the list below.)
  
2. Enter a succinct description that says why the PR is necessary, and what it does.
  - Implement aspect X
  - Leave out feature Y because of A
  - Improve performance by B
  - Improve accessibility by C

3. Provide clear testing instructions that include any pertinent information, i.e. seat, roles, etc.
4. Include screenshots of your changes if they impact the UI (Before & After).
5. Update the preview link with your pull request number.
6. Include any JIRA tickets, design documents, GitHub issues, or other related items to the bottom of the PR.

# Available labels
UI
API
Evaluation Server
OLAP

### DIFF
--- a/modules/back-end/src/Infrastructure/Webhooks/WebhookSender.cs
+++ b/modules/back-end/src/Infrastructure/Webhooks/WebhookSender.cs
@@ -7,6 +7,7 @@ using Domain.Webhooks;
 using HandlebarsDotNet;
 using Microsoft.Extensions.Logging;
 using System.Text.RegularExpressions;
+using Domain.Utils;
 
 namespace Infrastructure.Webhooks;
 
@@ -57,10 +58,7 @@ public partial class WebhookSender : IWebhookSender
             return delivery;
         }
 
-        var shouldPreventPayload = webhook.PreventEmptyPayloads
-                                   && jsonDocument.RootElement.ValueKind == JsonValueKind.Object
-                                   && !jsonDocument.RootElement.EnumerateObject().Any();
-        if (shouldPreventPayload)
+        if (webhook.PreventEmptyPayloads && jsonDocument.RootElement.IsEmptyObject())
         {
             return await PreventEmptyPayload(webhook.Id, events, webhook.PayloadTemplate, dataObject);
         }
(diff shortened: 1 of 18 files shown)


modules/back-end/src/Infrastructure/Webhooks/WebhookSender.cs
```diff
@@ -164,6 +177,22 @@ string ToHexString(IReadOnlyCollection<byte> bytes)
         }
     }
 
+    private async Task<WebhookDelivery> PreventEmptyPayload(Guid id, string events, string? payloadTemplate = default, Dictionary<string, object>? dataObject = default)
```
deleteLater: I'd prefer not to treat this as an error. So I made 2a7a6f7

drewsteinacher: Looks great! I was on the fence about this aspect as well, but wasn't sure how to proceed. Thanks!

deleteLater (approved):
LGTM, thanks for your time on this PR ❤️  @drewsteinacher

PR #682, opened 2024-11-15
✨ feat: validate insights data

wss-rbrennan:
This PR adds validation to several properties to avoid invalid data being processed or stored via the Track endpoint of the Insight Controller.  The following validations have been added.

| Class            | Property       | Validation Types |
|------------------|----------------|------------------|
| MetricInsight    | EventName      | Required, Regex  |
| VariationInsight | FeatureFlagKey | Required, Regex  |
| InsightMessage   | DistinctId     | Regex            |
| InsightMessage   | EnvId          | Regex            |
| Variation        | Id             | Regex            |

Unit tests have been added for these classes as well as unit tests for the InsightController.  The IsValid function has been expanded to validate MetricInsights, and Variations.  Some properties in the Insight object have been marked as virtual to allow for mocking in unit tests and an Interface IBoundedMemoryCache has been created for BoundedMemoryCache for the same reason.
  
To test this, beyond unit tests, you can use any of the sdks or curl to send a combination of valid or invalid requests. Note: if any insight is invalid in a batch that is flushed no insights from the batch will be processed.


deleteLater:
Hi @wss-rbrennan, I have made some refactoring to your code and I have some thoughts to share with you

1. We needn't to validate the `InsightMessage` instance because we instantiate it by using `MetricInsight` or `VariationInsight` instance, so we just need to ensure that `MetricInsight` and `VariationInsight` are valid.
2. About the variation id validation, I prefer to use `Guid.TryParse` instead of using regex.
3. For regex validation, I prefer to use `GeneratedRegex` attribute: https://learn.microsoft.com/en-us/dotnet/standard/base-types/regular-expression-source-generators.

But I still have a question about this PR. This PR's goal is to protect the public `/track` API, right? But I don't think that we can achieve this by simply adding some parameter validation, and maybe we should try something else.

wss-rbrennan:
The refactors look good. To respond to your points.

1. Agreed, though I was going towards a "validate at all levels" approach.
2. That works, also easier to read
3. I wasn't aware of that feature. thanks.

Overall the goal is to protect `/track` and prevent the application from handling or storing any invalid data (especially malicious data). I'm uncertain what additional approachs can be used other than validating the inputs.  The endpoint needs to remain available to all clients public and private, authentiated and unauthenticated.

PR #684, opened 2024-11-16
✨ feat: allow to set default permissions for organization

cosmos-explorer:
Allow to set default permissions for organizations, this would make it possible to set a customized permissions instead of current `Developer` permission

image


deleteLater (approved):
LGTM

PR #685, opened 2024-12-02
🐛 fix: end user insight search not working

joey5403:
🚀 change default period from 7D to 30M in Insight page to speedup request.
🐛 fix: End user search return empty list in Feature Flags -> Insights

This BUG only affects **FeatBit Standard**


deleteLater:
Thanks for finding this ☕

deleteLater (approved):
LGTM

PR #688, opened 2024-12-16
✨Sort Projects Alphabetically in Project/Env Modal

wss-alubic:
We have a lot of projects in FeatBit.  It's difficult to find the correct project in the list and it would be helpful to have the projects sorted alphabetically by name.


cosmos-explorer:
@wss-alubic I changed a the implementation, we think it would be better to sort the result on the backend

cosmos-explorer (dismissed):

deleteLater:
I prefer to use `localeCompare` because this method uses the client's default locale to compare string

wss-alubic:
Thanks everyone!

PR #714, opened 2025-03-04
🐛 fix: totalCount should never be None

cosmos-explorer:


modules/data-analytics/app/setting.py
```diff
@@ -85,7 +85,7 @@
         'formatter': 'default'
     }},
     'root': {
-        'level': 'INFO',
+        'level': 'WARN',
```
deleteLater: Wrong change

deleteLater (approved):
LGTM

PR #805, opened 2025-10-24
🐛 fix: 404 problem if BASE_HREF is set

cosmos-explorer:
If `BASE_HREF `is set, swithing project or environment would redirect to 404


deleteLater:
related to #796

PR #810, opened 2025-11-18
🏗 infra: migrate to bitnami legacy images

deleteLater:
This is a stopgap measure. The long-term goal is to move away from Bitnami entirely.

ref: https://github.com/bitnami/containers/issues/83267


cosmos-explorer (approved):
LGTM

PR #821, opened 2025-12-30
✨ feat: fine-grained access control for flags

cosmos-explorer:
This is the first implementation of  the fine-grained access control feature, it allows to configure different operations on feature flag. A later PR will be created for segments.

**Migration scripts**

**Postgresql**
```sql
UPDATE policies
SET statements = (
    SELECT jsonb_agg(
        CASE
            WHEN stmt->>'resourceType' = 'flag'
                 AND stmt->'actions' ? 'ManageFeatureFlag'
            THEN
                jsonb_set(
                    stmt,
                    '{actions}',
                    '["*"]'::jsonb,
                    false
                )
            ELSE
                stmt
        END
    )
    FROM jsonb_array_elements(statements) AS stmt
)
WHERE statements @> '[{"resourceType":"flag"}]';

UPDATE access_tokens
SET permissions = (
    SELECT jsonb_agg(
        CASE
            WHEN stmt->>'resourceType' = 'flag'
                 AND stmt->'actions' ? 'ManageFeatureFlag'
            THEN
                jsonb_set(
                    stmt,
                    '{actions}',
                    '["*"]'::jsonb,
                    false
                )
            ELSE
                stmt
        END
    )
    FROM jsonb_array_elements(permissions) AS stmt
)
WHERE permissions @> '[{"resourceType":"flag"}]';
```

**MongoDB**
```javascript
db.Policies.updateMany(
  {
    "statements.resourceType": "flag",
    "statements.actions": "ManageFeatureFlag"
  },
  {
    $set: {
      "statements.$[stmt].actions": ["*"]
    }
  },
  {
    arrayFilters: [
      {
        "stmt.resourceType": "flag",
        "stmt.actions": "ManageFeatureFlag"
      }
    ]
  }
)

db.AccessTokens.updateMany(
  {
    "permissions.resourceType": "flag",
    "permissions.actions": "ManageFeatureFlag"
  },
  {
    $set: {
      "permissions.$[perm].actions": ["*"]
    }
  },
  {
    arrayFilters: [
      {
        "perm.resourceType": "flag",
        "perm.actions": "ManageFeatureFlag"
      }
    ]
  }
);

```

PR #822, opened 2025-12-31
🐛 fix: access token drawer backward compatible

cosmos-explorer:
When the resouce type is flag, and the action is '*', the resource type would not be displayed in the edit drawer, current PR would display all available actions in this case
image


deleteLater:
related to #821

deleteLater (approved):
LGTM

PR #850, opened 2026-02-03
✨ feat: flag concurrent edits detection

deleteLater:
detect flag concurrent edits

## EN

image

## ZH

image


deleteLater:
We can do more on this problem by patch updating the flag. Currently this is the simplest way

PR #866, opened 2026-03-09
🐛 fix: guard message.data in websocket handler

Ocheretovich:
Prevent potential runtime error when websocket messages do not include the data field.


Ocheretovich:
Thanks! Removed the redundant guard

PR #883, opened 2026-03-22
🚀 : Cursor-based Pagination for End Users

wss-cadenwheeler:
## Migrate End Users Pagination from Offset-Based to Cursor-Based

This PR replaces offset-based pagination on the End Users page with cursor-based pagination to address
performance degradation at scale. Offset pagination requires the database to scan and skip an increasing
number of rows with each page, making it untenable when the end user count reaches into the millions.
Cursor-based pagination eliminates this overhead by anchoring each query to a specific record, resulting
in consistent query performance regardless of dataset size — validated to perform reliably at 50M+ end users.

### Also included in this PR

New indexes for both MongoDB and PostgreSQL to further improve query performance. The indexing strategies
are largely mirrored across both databases, with one notable addition for PostgreSQL: a GIN index using
trigram (`pg_trgm`) similarity, which dramatically improves partial-match and fuzzy search performance
across end user name fields, as well as segment and feature flag targeting queries.

### Summary of Changes

- Cursor-based pagination for the End Users page, performant at 50M+ end users
- New indexes for MongoDB and PostgreSQL
- PostgreSQL GIN/trigram index for name, key ID, segment targeting, and feature flag targeting searches

Screenshot of changes:

FeatbitCursorPagination

### Testing Steps

1. Populate MongoDB or PostgreSQL with 5M+ end user records.
2. Navigate to the End Users page.
3. Verify that pagination works correctly and performs consistently across pages.
4. **PostgreSQL only:** Verify that searching by name returns results with noticeably improved performance.

### Notes

> ⚠️ **Known MongoDB Limitation**

While cursor-based pagination improves end user page performance on both MongoDB and PostgreSQL,
improving the performance of name filtering on MongoDB remains an open problem. Unlike PostgreSQL,
MongoDB does not provide a mechanism to efficiently accelerate substring queries, and no viable
solution has been identified at this time.


modules/front-end/src/app/core/components/target-user/target-user.component.ts
```diff
@@ -102,8 +102,7 @@ export class TargetUserComponent implements OnInit {
       .filter(x => x.name.includes(value) || x.keyId.includes(value))
       .map(x => x.keyId);
 
-    // by default, we search for both env & global users
-    const filter = new EnvUserFilter(value, [], excludedKeyIds, true, false, 1, 5);
+    const filter = new EnvUserFilter(value, [], excludedKeyIds, false, false, 1, 5);
```
wss-cadenwheeler, line 105: Without this change, improving the performance of this query in its current state would be impossible. When `includeGlobalUser` is set to true, the query becomes an `OR` filter, which cannot be indexed in a single query. While including global users allows searching across workspaces, the performance implications make it better suited as an opt-in feature rather than opt-out.

deleteLater: Resolved in #915 by introducing a new `search` endpoint. The `target-user` component must support searches for both local and global users.

To optimize the performance of the `OR` filter query, we split it into two separate **indexed queries** combined into a single database round trip (`UNION ALL` for Postgres, `$unionWith` for MongoDb).

deleteLater:
done in #915

PR #899, opened 2026-05-18
✨ feat: added support and doc links

cosmos-explorer:
- As we now have the support tool available, so we replaced the feedback form by support link.
- We also added the link to our documentation site

image


deleteLater (approved):
LGTM

PR #911, opened 2026-06-01
✨ feat: allow resubscribe to current local subscription

cosmos-explorer:
If current user's subscription is created manually with the support team (done by payment link), we should allow the user to renew the subscription from the UI


deleteLater (approved):
LGTM

PR #931, opened 2026-06-24
↩️ revert: restore project switcher modal behavior

wss-cadenwheeler:
## Summary
 
 The project selector button in the global header (`header.component.html`)
 was navigating to `/organization/projects` instead of opening the
 project/environment switcher modal. This PR restores the original
 in-place modal behavior so the button matches its sibling environment
 button and lets users switch context without leaving the page they're on.
 
 ## Change
 
 ```diff
 - <div class="selector-item" [routerLink]="['/organization/projects']"
 + <div class="selector-item" (click)="envModalVisible=true"
```

Why

 1. Lost context. Routing to /organization/projects drops the user
 out of whatever flag/segment/experiment they were working in. The modal
 keeps them in place.
 2. Inconsistency. The two adjacent selector items (project + env) had
 different interaction models for what is conceptually the same action
 ("switch the scope I'm working in"). Users expected them to behave the
 same way.
 3. Regression. This was the prior behavior; the routerLink was
 introduced later and never replaced the modal-based UX.
4. Avoids a performance pitfall. The Projects page degrades sharply with project count: organizations with large project lists routinely see 15+ second UI freezes, and in some cases the page fails to load at all. Keeping users in the in-place modal sidesteps that path entirely.

Testing

 - Verified locally that clicking the project button now opens the
 existing environment/project switcher modal.
 - Verified the environment button still works (unchanged).
 - Verified the modal's "switch" action still navigates correctly and
 closes the modal (existing behavior in header.component.ts).
 - No new components, services, or styles introduced.

### DIFF
--- a/modules/front-end/src/app/core/components/header/header.component.html
+++ b/modules/front-end/src/app/core/components/header/header.component.html
@@ -63,7 +63,7 @@
             </div>
           </div>
           <div class="selector-divider"></div>
-          <div class="selector-item" [routerLink]="['/organization/projects']"
+          <div class="selector-item" (click)="envModalVisible=true"
                [nz-tooltip]="isSmallScreen ? projectTooltip : null"
                nzTooltipPlacement="bottom">
             <div class="item-icon proj-icon">


deleteLater:
> Avoids a performance pitfall. The Projects page degrades sharply with project count: organizations with large project lists routinely see 15+ second UI freezes, and in some cases the page fails to load at all. Keeping users in the in-place modal sidesteps that path entirely.

~~We need to address the performance issues in a separate PR.~~

#935

deleteLater (approved):
LGTM

PR #940, opened 2026-07-09
✨ feat: control plane module with cross-DC gated-commit consistency

wss-rbrennan:
## What this adds

An optional **Control Plane** service — a new .NET 10 module (`modules/control-plane`) that
brokers flag/segment/secret changes between the API server and evaluation servers across
multiple regions/data centers, plus an opt-in **gated-commit consistency mode**: when
enabled, no flag or segment version is served by any live DC until every live DC has it
(stage → gate on per-DC probes and heartbeat leases → version-guarded commit →
pointer-gated eval reads → dead-DC eviction → recovery backfill and cache self-heal →
readiness fence).

## Why

FeatBit today has no first-party story for running one logical instance across multiple
regions/DCs — each DC's evaluation servers only see changes that reach their local Redis and
message queue. The control plane provides that brokering: a flag/segment/secret change made
through the API server is fanned out to every DC's Redis and republished to that DC's
evaluation servers.

In the default **BestEffort** mode the fan-out is fire-and-forget: if one DC's Redis is
unreachable or a whole DC is down, the change is served by the healthy DCs while the lagging
one still serves the old value — SDK clients in different DCs see different values.
**GatedCommit** closes that window with the invariant above — without Redis Enterprise or
CRDTs, using the existing per-DC Redis plus a commit coordinator gated on the source of truth
(Mongo/Postgres).

Everything is **opt-in and off by default**: the control plane only engages when the API server
runs with `UseControlPlane=true` — single-cluster deployments run unchanged code paths — and
consistency defaults to `ConsistencyMode=BestEffort`. The module follows the same conventions
as the existing services: Clean Architecture (`Api → Application → Domain → Infrastructure`),
pluggable providers (`DbProvider`: MongoDB/Postgres, `MqProvider`: Kafka/Redis/Postgres), and
`/health/liveness` + `/health/readiness` endpoints. Storage additions are additive only
(`dc_leases`, pending-change records); no migrations touch existing collections/tables.

## Design

The control plane gates commits on its own per-DC Redis writes + lease liveness (it does not
depend on the evaluation servers to confirm). Lifecycle, per change, under GatedCommit:

1. **Stage** — write the new value into every DC's Redis under a versioned key
   (`…:{id}:v{ts}`) and record a pending change in the DB; withhold the eval-server publish.
2. **Gate** — a commit coordinator waits until every live DC (a DcId with ≥1 unexpired
   pod-heartbeat lease) actually has `v{ts}` (probed directly in each DC's Redis).
3. **Commit** — advance the committed pointer + index in every DC (guarded so they can only move
   forward), promote pending→committed in the DB (version-guarded, atomic on Mongo, `xmin`
   optimistic-concurrency with bounded retry on Postgres), then republish to the eval servers.
4. **Serve** — eval servers read the committed version from their local Redis (pointer-gated
   reads), so full-sync / reconnect / incremental-sync never serve an uncommitted value.
5. **Evict** — a DC that stops heartbeating is dropped from the live set after its lease TTL (so a
   dead DC can't block commits forever); commit proceeds on the rest.
6. **Recover** — when a DC returns, a recovery worker backfills its Redis to the current committed
   state (flags, segments, and secrets) and refreshes its connected clients before it rejoins.
   A mode-agnostic cache reconciler additionally self-heals any DC's cache from the source of
   truth whenever its Redis link becomes reachable.
7. **Fence** — under GatedCommit, an eval server that can't heartbeat past a threshold (likely
   evicted/partitioned) fails `/health/readiness` and is pulled from rotation (strict
   consistency-over-availability; readiness-only, returns when heartbeats resume).

A partitioned-but-alive DC's committed pointer cannot advance, so its reads freeze at the last
committed value (consistent-but-stale) until it converges on return. The commit flip itself is
bounded-skew by documented design (normally sub-second): by the time pointers advance, every
live DC already holds the new version, so the residual skew is only "old committed vs new
committed", never "a value a DC doesn't have".

## What's included (by area)

- **Versioned storage** — committed-vs-pending model on FeatureFlag/Segment (Mongo +
  Postgres); staged/committed Redis key scheme with only-advance guards (Lua compare-and-set +
  `ZADD GT`) so no backfill or race can ever move a served value backward; staged-version GC;
  Postgres `xmin` as an optimistic concurrency token with a shared, jittered-backoff retry policy
  and loud exhaustion logging.
- **Membership** — `dc_leases` store (Mongo + Postgres); eval-server heartbeats (5s default,
  coherent with the 15s lease TTL) carry DcId/Region and a per-environment applied watermark
  (newest committed flag/segment the DC serves); a cadence warning fires when heartbeats arrive
  slower than the lease TTL.
- **Coordinator & workers** — DcId-labeled composite cache with per-DC staged probes; the commit
  coordinator (flags + segments, monotonicity + optimistic-promote guards, eviction logging);
  recovery worker sharing one committed snapshot per tick; the cache self-heal reconciler
  (reconnect cooldown, cross-worker coalescing, accepted-vs-attempted repair accounting); an
  advisory DcId-mismatch checker.
- **Leader election (opt-in, default OFF)** — a Redis TTL-lock elector can restrict the
  consistency workers to one control-plane replica. Disabled by default: single-replica
  deployments gain nothing and would inherit a stall surface; without election, multi-replica is
  safe-but-redundant by construction (all operations are idempotent and version-guarded). Segment
  changes preserve the original operator attribution through the gated pipeline.
- **Serving** — pointer-gated eval-server reads (flags + segments); gated change handlers
  stage-and-hold; self-healing Redis connections everywhere (`abortConnect=false` + bounded
  timeouts — a failed first connect is no longer cached for the pod's lifetime).
- **Observability** — `FeatBit.ControlPlane.Consistency` and eval-side meters: commits,
  time-to-commit, pending backlog, evicted commits, per-DC applied-watermark lag (frontier = max
  across live DCs), is_leader, unmatched-DC count, heartbeat staleness; gauge snapshots reset on
  non-leader replicas so stale values are never exported; per-DC Kafka consumer group ids (derived
  from the local DcId) so control planes sharing a broker never silently split the change stream.
- **QA harness** — a two-cluster (west/east) validation environment: per-cluster Redis+Sentinel
  with a cross-cluster master-forwarder, chaos-mesh partitions, an automated 15-scenario suite
  (`cp01–cp15`: propagation correctness, resilience under Redis outage, environment/secret/license
  propagation, pod-failover with WebSocket client migration, and the consistency set — happy path,
  eviction, recovery, readiness fence, mode toggle/rollback, cache self-heal), an
  OpenTelemetry-Demo application instrumented with FeatBit SDKs across 10 services
  (Apache-2.0 attribution preserved and documented), and provenance-verified image builds.
- **Docs** — module READMEs summarize the feature and link to the canonical operator guide
  (`control-plane-qa/00-Docs/GATED-COMMIT-CONSISTENCY.md`): mechanism, full config reference,
  metrics, rollout/rollback procedure, validation checklist, known limitations, and the recorded
  design decisions (bounded-skew vs linearizable reads; opt-in leader election; the reconciler
  deliberately ungated).

## Enabling (default OFF)

- Control plane + every evaluation server: `ControlPlane:ConsistencyMode=GatedCommit`
  (default `BestEffort` = the plain fire-and-forget fan-out described above).
- Each control-plane `Redis:Instances[]` entry gets a `DcId`; each eval server gets a matching
  `ControlPlane:DcId` — these must match (the coordinator's join key; a mismatch is surfaced
  by the advisory checker).
- Tunables (sensible defaults, full reference in the operator guide): `LeaseTtlSeconds`,
  `HeartbeatIntervalSeconds`, `HeartbeatStalenessThresholdSeconds`, worker intervals,
  `LeaderElection:{Enabled,TtlSeconds,RenewIntervalSeconds}`, `CacheReconcile:*`,
  Redis connect/sync timeouts.

## Verification

- **Test suites:** 1,607 tests green across the three touched modules (Release configuration),
  including the Testcontainers-backed integration suites against real Mongo/Redis/Postgres/Kafka
  (these are intentionally skipped in CI, which filters `Category!=Integration`, but still build
  there; locally, Docker is the only prerequisite — they self-skip without it). The control-plane
  module is covered by the testing standard in `CONTRIBUTING-tests.md` and ships its own CI
  workflow (`build-and-test-control-plane.yml`) with coverage artifacts, matching the other
  modules.
- **Adversarial review:** the branch went through three multi-agent review passes (independent
  finder angles + per-finding verification). Every confirmed finding was fixed in-branch —
  including a corruption-class gap in an early version of the only-advance guard (single-env gate
  on multi-env segments) and a readiness-fence default left incoherent by the heartbeat-interval
  change. Refuted candidates and accepted trade-offs are documented in code comments and the
  operator guide.
- **Live validation:** the **full cp01–cp15 suite passes** against a from-scratch two-cluster
  GatedCommit deployment on images built from this branch's tip and provenance-verified (the
  build script stamps each image with its source commit after an earlier stale-image trap was
  found and fixed). That includes real network-partition chaos (DC eviction + recovery +
  backfill + readiness fence), WebSocket client failover (clients migrate across DCs through an
  active/active LB when a DC's eval pods die — verified on k6 0.57), and the complete
  mode flip cycle with real config rollouts: BestEffort → GatedCommit → **rollback**, all green.
  Dated validation reports, including an erratum for an earlier stale-image run and
  cluster-recovery troubleshooting, are in `control-plane-qa/02-Tests/simulations/`.

## Rollback

Set `ConsistencyMode=BestEffort` and restart — no data migration; staged `v{ts}` keys are harmless
and GC'd. BestEffort serving paths remain unchanged (the self-heal only adds recovery writes,
which are version-guarded). The rollback path is exercised live by the mode-toggle scenario.

## Deferred follow-ups (intentionally out of scope)

- `CacheReconciler` is deliberately not leader-gated (runs in both modes; documented).

## Size and composition

| Area | Lines | Share |
|---|---|---|
| `control-plane-qa/` (test harness, deployment scripts, docs) | 47,021 | 64.0% |
| `modules/` — test projects | 13,168 | 17.9% |
| `modules/` — application source | 9,394 | 12.8% |
| other (docker, docs, CI, root) | 1,568 | 2.1% |
| `kubernetes/` manifests | 1,142 | 1.6% |
| `benchmark/` k6 scripts | 1,191 | 1.6% |
| **Total** | **73,484** | |

The actual product change is deliberately small — ~9.4k lines of application source (the new
module plus conditional hooks in the API and evaluation servers). The bulk is the verification
apparatus built alongside it; ~87% of the diff is test/QA/docs content that cannot affect
production behavior.

## Reviewing

Given the size, review by area rather than by commit. Suggested reading order — start with the
operator guide (`control-plane-qa/00-Docs/GATED-COMMIT-CONSISTENCY.md`) as the map, then:

1. **Domain & storage** — the committed-vs-pending model and its guards:
   `modules/back-end/src/Domain/{FeatureFlags,Segments}/`,
   `modules/back-end/src/Infrastructure/Services/{MongoDb,EntityFrameworkCore}/`.
2. **Redis cache layer** — staged/committed keys, only-advance Lua scripts, self-healing
   connections: `modules/back-end/src/Infrastructure/Caches/Redis/`.
3. **Control-plane workers** — coordinator, recovery, reconciler/backfiller, DcId checker,
   leader election: `modules/control-plane/src/Api/Application/ControlPlane/`.
4. **Multi-DC composite cache & wiring** — per-DC clients, staged probes, targeted writes,
   consumer-group derivation: `modules/control-plane/src/Api/Infrastructure/{Caches,MQ}/`.
5. **Evaluation server** — pointer-gated reads, heartbeats + applied watermark, readiness fence:
   `modules/evaluation-server/src/{Infrastructure/Store,Streaming/Health,Api/Health}/`.
6. **QA harness & scenarios** — `control-plane-qa/` (deploy scripts, chaos manifests,
   `02-Tests/automation-py/scenarios/`).

Each area's tests sit alongside it (unit projects are infra-free; everything infra-dependent
lives in the Testcontainers-backed integration projects).

If the maintainers prefer, we're happy to split this into review-sized stages (module skeleton →
consistency machinery → QA suite) or to start with a feature-suggestion Discussion thread per
CONTRIBUTING.md — whichever fits your workflow best.


deleteLater:
**Big changes made**

- 863a03b, move `control-plane-qa` to `e2e/control-plane` and update related paths. After that, tested via `poetry run automation suite cp02 --seed-data` against `Basic` deployment mode.

**Migration Script**

Will be added in a v6.0.0 preview release note

```sql
\connect featbit

-- Control plane data center membership leases (Option A consistency).
-- One row per data center, keyed by dc_id. Mirrors the Mongo "dc_leases" collection.
CREATE TABLE IF NOT EXISTS dc_leases
(
    dc_id              text PRIMARY KEY,
    region             text,
    last_heartbeat_at  timestamptz,
    lease_expires_at   timestamptz,
    applied_watermarks jsonb
);

-- The live-set query filters on lease_expires_at > now().
CREATE INDEX IF NOT EXISTS ix_dc_leases_lease_expires_at
    ON dc_leases (lease_expires_at);

-- B4: committed-vs-pending for feature_flags (Postgres parity with the Mongo path).
-- committed_version: monotonic version of the last COMMITTED change (default 0).
-- pending: a staged-but-not-committed change, stored as jsonb (NULL when none).
ALTER TABLE feature_flags
    ADD COLUMN committed_version bigint NOT NULL DEFAULT 0,
    ADD COLUMN pending jsonb NULL;

-- S1: committed-vs-pending for segments (Postgres parity with the Mongo path).
-- committed_version: monotonic version of the last COMMITTED change (default 0).
-- pending: a staged-but-not-committed change, stored as jsonb (NULL when none).
ALTER TABLE segments
    ADD COLUMN committed_version bigint NOT NULL DEFAULT 0,
    ADD COLUMN pending jsonb NULL;
```

**Memo/Reminder**

- test scripts repo setup (clone from https://github.com/wss-cadenwheeler/featbit then switch to `control-plane` branch

**TO CONFIRM**
- [ ] `ConnectionMessage.Secret` field is missleading, its actual value is `projectKey`. As a result, `uat-connections.js` has the following missleading code
	```js
	(c) => (envId && c.envId === envId) || (projectKey && c.secret === projectKey)
	```
- [ ] When controlplane enabled, we fires webhook at *stage* time rather than *commit* time. If that's what we want, it seems completely unneccessary, we can just do that on the API side. Or what we want to archive is "fire only after cross-DC sync confirmed" semantics"?
- [ ] No GC for staged segment versions — `StagedFlagGcWorker.cs`, couldn't find any equivalent sweep for `featbit:segment:*:v*` anywhere
- [ ] **API Service**: Migrate `UseControlPlane` to `ControlPlane:Enabled` for consistency
- [x] **BUG:** Every time a WebSocket connection is added or removed, we unconditionally publish a message to the message queue. Only the `ControlPlane` can consume these messages, but the `ControlPlane` is an optional component. fixed in 4b967cf
- [ ] Docs about Data Flows between **API, Control Plane and Evaluation Server**
- [ ] Docs about Control Plane architecture

wss-rbrennan:
> * [ ]  When controlplane enabled, we fires webhook at _stage_ time rather than _commit_ time. If that's what we want, it seems completely unneccessary, we can just do that on the API side. Or what we want to archive is "fire only after cross-DC sync confirmed" semantics"?


When controlplane enabled and in GatedCommit mode, webhooks should only fire after cross-DC sync. Otherwise consumers may be informed of a change that is not yet live.

deleteLater:
> When controlplane enabled and in GatedCommit mode, webhooks should only fire after cross-DC sync. Otherwise consumers may be informed of a change that is not yet live.

The current implementation does not satisfy that requirement; we can implement this feature in a future PR.

This PR already contains too much, so the items on the list above should be addressed in separate, future PRs.

deleteLater:
# Code Review On `modules/control-plane/src/Api`

***This is a LLM generated code review report; manual verification is required before making any changes.***

## High severity

**1. `app.Run()` is called twice** _fixed in 05c992c_
MiddlewareRegister.cs ends with `app.Run(); return app;`, and Program.cs does `.SetupMiddleware().Run()`. The first `Run()` blocks until shutdown; the second then tries to start an already-stopped host. It throws on every graceful shutdown (masked because the process is exiting anyway). One of the two calls should go.

**2. Connection handlers write to the local DC only, but deletes broadcast to all DCs**
Every message handler injects `[FromKeyedServices("compositeCache")] ICacheService` — except ClientConnectionMadeHandler.cs and ClientConnectionClosedHandler.cs, which take the plain `ICacheService`. That resolves to `RedisCacheService` over the singleton `IRedisClient`, i.e. `Redis:Instances[0]` — the local DC only.

Meanwhile CompositeRedisCacheService.cs broadcasts `DeletePodConnection`, justified by a comment that is now factually wrong:

> `DeletePodConnection still broadcasts, because it also clears the pod's connection keys, and those were broadcast on connect by UpsertConnectionMadeAsync.`

They are not broadcast on connect. Either the two handlers are missing the keyed attribute, or the delete should be local-only and the comment corrected. Right now the write and delete paths have different DC scope.

**3. `RecoveryWorker` permanently forgets a DC if its backfill throws**
RecoveryWorker.cs advances `_previousLiveSet = liveDcs;` *before* the backfill loop. The composite-cache-unavailable path explicitly un-advances it, but if `FetchCommittedSnapshotAsync` or `BackfillDcAsync` throws, the exception escapes to the tick's `catch` with the watermark already advanced. That DC is no longer "newly present" on any future tick, so it is never backfilled again — it stays stale until the control plane restarts or the DC's lease lapses and returns. CacheReconciler.cs gets this right (`_lastConnected.Remove(dcId)` in its catch); `RecoveryWorker` needs the equivalent try/catch.

**4. Startup crash when `CacheProvider=Redis` but `Redis:Instances` is empty**
CacheServiceCollectionExtensions.cs does `services.AddSingleton<IRedisClient>(clients[0]);` after `?? []`. An empty or missing `Redis:Instances` gives an unhandled `ArgumentOutOfRangeException` at registration time with no diagnostic. Worth a guard with a clear message, since every other misconfiguration in this module degrades gracefully.

**5. `RedisLeaderElector` locks itself out for a full TTL after a transient blip**
RedisLeaderElector.cs: any Redis exception sets `_isLeader = false`. But the lock key is still in Redis holding *this* instance's value. The next tick takes the `!_isLeader` branch → `LockTakeAsync` (SET NX) → fails because the key exists. So the instance cannot recover leadership until the TTL (default 30s) expires, and no other replica can take it either. A single 10ms network hiccup produces up to 30s with **no leader**, during which `CommitCoordinatorWorker`, `RecoveryWorker` and `DcIdConsistencyChecker` all skip every tick. Fix: attempt `LockExtend` first and fall back to `LockTake`, or `LockRelease` before dropping the flag.

**6. Handler exceptions lose messages on Kafka and Redis MQ**
The change handlers deliberately rethrow after logging (e.g. FeatureFlagChangeMessageHandler.cs), which only produces a retry under the Postgres MQ. In `KafkaMessageConsumer` the offset is stored in a `finally` block that runs even after the handler threw, and `RedisMessageConsumer` has already `LPOP`ed the message. Under `MqProvider=Kafka` (the pro edition), a transient Redis blip during staging silently drops a flag change permanently. The rethrow reads as "this will be retried" — on two of three providers it isn't.

## Medium severity

**7. Unknown live DC stalls *all* commits globally**
CommitCoordinatorWorker.cs: `liveDcs.All(dc => stagedMap.TryGetValue(dc, out var staged) && staged)`. `stagedMap` keys come from configured `Redis:Instances`. An ELS pod reporting a `DcId` with no matching Redis instance is in `liveDcs` but never in `stagedMap`, so the gate can never be satisfied — for *every* flag and segment, not just that DC. DcIdConsistencyChecker.cs correctly diagnoses this but is advisory-only. A single typo in one ELS's `ControlPlane:DcId` silently freezes the entire commit pipeline. This deserves at least a `LogWarning` at the gate itself, not just from a separate 60s worker.

**8. Lease expiry depends on the ELS pod's clock**
HeartbeatMessageHandler.cs sets `LeaseExpiresAt = heartBeatMessage.Timestamp + leaseTtl` using the *pod's* timestamp, but `GetLiveSetAsync(DateTimeOffset.UtcNow)` compares against the *control plane's* clock. Skew greater than the 15s TTL means a DC is either never live (commits stall forever) or always live (a dead DC is gated on forever). Using `DateTimeOffset.UtcNow + leaseTtl` on the control plane side would remove the cross-host clock dependency; `LastHeartbeatAt` can keep the reported value.

**9. `DeletePodConnection` is O(all connections) and broadcast per dead pod**
`PodHealthChecker` calls `cacheService.DeletePodConnection(podId)` per unhealthy pod → composite broadcast → per DC, `RedisCacheService.DeletePodClients` runs a full `SCAN featbit:connection:*` plus an `HGET` per key. With 100k SDK connections, 5 dead pods and 2 DCs that's 10 full keyspace scans and ~1M round trips per health tick. A `heartbeatId → connectionIds` set would make this O(connections-for-that-pod).

**10. `GET /api/admin/connections` returns SDK secrets in plaintext**
GetConnectionsHandler.cs projects `secret` straight into the response. Combined with a single static API key (finding 12), that's a full credential dump from one endpoint. Also: `GetServers().FirstOrDefault(s => s.IsConnected)` scans only one node (wrong under Redis Cluster), and reads only the local DC. Consider dropping/masking `Secret` unless something genuinely consumes it.

**11. Non-constant-time API key comparison**
ApiKeyAuthenticationHandler.cs uses `_apiKey.Equals(providedApiKey, StringComparison.OrdinalIgnoreCase)`. `String.Equals` short-circuits on first difference — a timing oracle. `OrdinalIgnoreCase` also throws away roughly a bit per character of a hex/GUID key. Use `CryptographicOperations.FixedTimeEquals` over the UTF-8 bytes with ordinal semantics.

**12. Timezone-sensitive version tokens**
Version tokens are `new DateTimeOffset(flag.UpdatedAt).ToUnixTimeMilliseconds()` in FeatureFlagChangeMessageHandler.cs, SegmentChangeMessageHandler.cs and DcBackfiller.cs. `new DateTimeOffset(DateTime)` applies the **local** offset when `Kind == Unspecified`. The handler's flag comes from JSON deserialization; the backfiller's comes from Mongo/EF. If those two paths ever differ in `Kind`, the same flag gets version tokens differing by the host's UTC offset, which breaks the only-advance guard and the staged-version gate in both directions. Note that CommitCoordinatorWorker.cs already defends against exactly this with `DateTime.SpecifyKind(resourceUpdatedAt, DateTimeKind.Utc)` — the inconsistency is the tell. Invisible in UTC containers, which is what makes it a good hidden bug. Normalizing in `DateTimeExtensions.ToUnixTimeMilliseconds` would cover all call sites.

**13. `CacheReconciler` can miss a Redis restart**
Detection is a 10s poll of `IsConnected`. A Redis that restarts empty and reconnects inside one poll interval never produces an observed `false`, so no transition fires and the cache stays empty. Under `BestEffort` nothing else repairs it (`RecoveryWorker` is GatedCommit-only). A cheap belt-and-braces addition would be checking the `featbit:redis-is-populated` marker, or subscribing to `ConnectionRestored` on the multiplexer instead of polling.

**14. Unbounded full-table scans on a 5s timer**
`GetPendingAsync()` runs every coordinator tick and `GetAllCommittedAsync()` on every backfill — both `_ => true`-style scans over all flags/segments across all workspaces, with no paging. `DcBackfiller.RunBackfillAsync` then awaits one write per flag, per segment, per secret, sequentially. On a large install a single backfill is minutes of serialized round trips while holding the DC's in-flight slot.

## Low / nits

- MiddlewareRegister.cs: `UseHttpsRedirection()` is registered *after* `UseAuthentication`/`UseAuthorization`, so the API key is parsed off a plaintext request before the redirect. Move it to the top of the pipeline.
- CacheServiceCollectionExtensions.cs: `connStr += $",password={instance.Password}"` — a password containing `,` or `=` corrupts the connection string. Same for the `abortConnect`/timeout appends if a password contains those tokens.
- SecretChangeMessageHandler.cs: `Enum.TryParse` accepts arbitrary numeric strings (`"7"` parses successfully as an undefined enum value). Add `Enum.IsDefined`, or the `_ =>` arm carries the load — it does here, but only by luck.
- FeatureFlagChangeMessageHandler.cs: `StageFlagAsync` returns `bool` and the result is discarded. Correct by design (the coordinator re-probes), but a `LogDebug` on a partial stage would make stalled commits far easier to diagnose.
- ApiControllerBase.cs: the base route is `api/v{version:apiVersion}/[controller]` but no API versioning is registered. `AdminController` overrides it, so nothing breaks today — but the next controller that inherits without its own `[Route]` will fail at route-construction time.
- Several DI comments in CommitCoordinatorWorker.cs and DcBackfiller.cs say these services are "scoped (per-request)", but DbServiceCollectionExtensions.cs registers them `AddTransient`. The scope-creation is still correct practice; the comments are just wrong.
- The gauge snapshots in `CommitCoordinatorWorker` and `DcIdConsistencyChecker` are `static`, so two instances in one process overwrite each other — the exact hazard `RedisLeaderElector` documents and avoids with an instance-owned `Meter`. Only affects tests today.

## Suggested priority

Fix 1–5 first (all are correctness bugs with concrete failure modes and small diffs). 7 and 8 are the ones most likely to produce a confusing production incident — a silently frozen commit pipeline with no error anywhere. 10 and 11 I'd raise with whoever owns the security review.

deleteLater:
_Copied from a previous comment (since it's collapsed by default on the webpage)_

**Semantic Changes I Made**

**ELS**

- 4b967cf publish connection-made/closed messages only when Control Plane is enabled
- cbe32cc fixed a RedisMessageConsumer bug
- 13f0011 use the pointer-aware store (GatedCommitRedisStore) only in GatedCommit mode
- ef63103 consume Topics.ControlPlaneCommand messages only when Control Plane is enabled
- 05c992c fixed app.Run() being called twice

**API**

- cba7213 – refactor the way we refresh access token "last used at" timestamp
- 9b01d1b – addressed PR feedback on PostgresLeaseStore.cs (logic fix)

**MISC**

- 863a03b, infra: move `control-plane-qa` to `e2e/control-plane` and update related paths. After that, tested via `poetry run automation suite cp02 --seed-data` against `Basic` deployment mode.

All other changes are either structural, non-functional refactors, test/infra/build fixes, or doc/comment changes.

**TODO**

- [ ] test scripts repo setup (clone from https://github.com/wss-cadenwheeler/featbit then switch to `control-plane` branch
- [ ] `ConnectionMessage.Secret` field is missleading, its actual value is `projectKey`. As a result, `uat-connections.js` has the following missleading code. This misleading field name (in my case) caused the LLM to make incorrect judgments several times.
	```js
	(c) => (envId && c.envId === envId) || (projectKey && c.secret === projectKey)
	```
- [ ] When controlplane enabled, we fires webhook at *stage* time rather than *commit* time. If that's what we want, it seems completely unneccessary, we can just do that on the API side. Or what we want to archive is "fire only after cross-DC sync confirmed" semantics"?
- [ ] No GC for staged segment versions — `StagedFlagGcWorker.cs`, couldn't find any equivalent sweep for `featbit:segment:*:v*` anywhere
- [ ] **API Service**: Migrate `UseControlPlane` to `ControlPlane:Enabled` for consistency
- [x] **BUG:** Every time a WebSocket connection is added or removed, we unconditionally publish a message to the message queue. Only the `ControlPlane` can consume these messages, but the `ControlPlane` is an optional component. fixed in 4b967cf
- [x] Rolling back from `GatedCommit` to `BestEffort` leaves flag and segment evaluation data stale because existing committed pointers continue to override newer legacy main-key values, so reads must honor pointers only in `GatedCommit` and transition tests should cover both values coexisting. fixed in 13f0011
	1. `GatedCommit` writes a versioned value and committed pointer.
	2. Deployment switches to `BestEffort`.
	3. `BestEffort` writes newer data only to the legacy main key in `RedisCacheService.cs`
	4. `RedisStore.cs` still sees the old pointer and unconditionally reads the old versioned value.
	5. Subsequent `BestEffort` updates remain invisible because nothing deletes or ignores that pointer.

**Docs**

- [ ] Staled docs about control plane topic literals — the real prefix is featbit-control-plane-*, not the cp-* implied by the repo instructions.
- [ ] Docs about Data Flows between **API, Control Plane and Evaluation Server**
- [ ] Docs about Control Plane architecture

deleteLater:
_Copied from a previous comment (since it's collapsed by default on the webpage)_

**Migration Script**

Will be added in v6.0.0 preview release note

```sql
\connect featbit

-- Control plane data center membership leases (Option A consistency).
-- One row per data center, keyed by dc_id. Mirrors the Mongo "dc_leases" collection.
CREATE TABLE IF NOT EXISTS dc_leases
(
    dc_id              text PRIMARY KEY,
    region             text,
    last_heartbeat_at  timestamptz,
    lease_expires_at   timestamptz,
    applied_watermarks jsonb
);

-- The live-set query filters on lease_expires_at > now().
CREATE INDEX IF NOT EXISTS ix_dc_leases_lease_expires_at
    ON dc_leases (lease_expires_at);

-- B4: committed-vs-pending for feature_flags (Postgres parity with the Mongo path).
-- committed_version: monotonic version of the last COMMITTED change (default 0).
-- pending: a staged-but-not-committed change, stored as jsonb (NULL when none).
ALTER TABLE feature_flags
    ADD COLUMN committed_version bigint NOT NULL DEFAULT 0,
    ADD COLUMN pending jsonb NULL;

-- S1: committed-vs-pending for segments (Postgres parity with the Mongo path).
-- committed_version: monotonic version of the last COMMITTED change (default 0).
-- pending: a staged-but-not-committed change, stored as jsonb (NULL when none).
ALTER TABLE segments
    ADD COLUMN committed_version bigint NOT NULL DEFAULT 0,
    ADD COLUMN pending jsonb NULL;
```

deleteLater:
Hello @wss-rbrennan, I’ve mostly finished reviewing this PR. 

I’ve gone through the changes in the `src` folders for the API and ELS services one by one, and I used the copilot to review the newly added ControlPlane module.

In [Comment 1](https://github.com/featbit/featbit/pull/940#issuecomment-5126834919) and [Comment 2](https://github.com/featbit/featbit/pull/940#issuecomment-5126887807), I’ve listed the to-do items and some bugs to be fixed regarding the `ControlPlane` module for your reference. If it's ok for you, I'll merge this PR tomorrow, and it will be included in the **v6.0.0-preview1** release.

Addtionally, thank you to your team for the time you’ve spent on this PR ❤️

wss-rbrennan:
@deleteLater Sounds good to me, let us know if you need assistance with the remaining issues [Comment 1](https://github.com/featbit/featbit/pull/940#issuecomment-5126834919) [Comment 2]( https://github.com/featbit/featbit/pull/940#issuecomment-5126887807)

deleteLater:
# General Review against PR description
 
(for memo and reference)
 
 ***This is a LLM generated code review report; manual verification is required before making any changes.***

1. **[P0] Partial Redis commit still promotes and publishes.**
`CommitCoordinatorWorker.cs:415` ignores the broadcast result. Per-DC failures are swallowed in `CompositeRedisCacheService.cs:410`. If one live DC fails after staging was probed, the DB is promoted and healthy DCs publish the new value while the failed DC remains old. The segment path has the same issue at `CommitCoordinatorWorker.cs:509`.
2. **[P0] PostgreSQL deployments never receive the required schema.**
The dc_leases, committed_version, and pending DDL exists only in the test fixture `vNext.sql:3`. Real deployments mount `docker-entrypoint-initdb.d` in `docker-compose-dev-controlplane.yml:121`, where no corresponding migration exists. A PostgreSQL GatedCommit deployment therefore fails when these tables/columns are accessed.
3. **[P1] Failed staging is acknowledged and can block a change indefinitely.**
The composite returns false when any DC fails, but both handlers discard it: `FeatureFlagChangeMessageHandler.cs:50` and `SegmentChangeMessageHandler.cs:57`. The pending DB record remains, the message is acknowledged, and no worker restages pending versions. The coordinator then waits forever for the missing staged key.
4. **[P1] A superseded pending version can become the served pointer.**
Redis pointers advance before the guarded DB promotion at `CommitCoordinatorWorker.cs:415`. If a newer pending version replaces the snapshot before `CommitCoordinatorWorker.cs:420`, promotion returns false but the old pointer remains live. Full sync and reconnect can then serve a version the source of truth never promoted.
5. **[P1] The Kubernetes readiness fence is ineffective.**
`evaluation-server-deployment.yaml:45` uses a TCP readiness probe. It remains successful when /health/readiness returns 503, so heartbeat fencing never removes the pod from Kubernetes service rotation.
6. **[P1] Following the SPEC’s enablement instructions does not enable gated eval behavior.**
The SPEC only requires ControlPlane:ConsistencyMode=GatedCommit, but registration is additionally guarded by ControlPlane:Enabled at `ServicesRegister.cs:79`, which defaults to false in `appsettings.json:81`. The existing test explicitly expects GatedCommit with Enabled=false to retain the legacy store.
7. **[P1] Readiness tracks broker publication, not successful lease renewal.**
Eval marks heartbeat success immediately after MQ publication at `HeartbeatService.cs:36-40`. The control plane catches and suppresses lease processing failures at `HeartbeatMessageHandler.cs:55`. A consumer or lease-store failure can therefore expire the DC lease while eval pods remain ready.

deleteLater:
> @deleteLater Sounds good to me, let us know if you need assistance with the remaining issues [Comment 1](https://github.com/featbit/featbit/pull/940#issuecomment-5126834919) [Comment 2](https://github.com/featbit/featbit/pull/940#issuecomment-5126887807)

My primary work after this will be on removing the existing `da-server` and introducing the new data experimentation module. Therefore, I may not be able to address these issues right away. If there are any high-priority issues that you’re concerned about, feel free to submit a fix.

deleteLater (approved):
LGTM

PR #921, opened 2026-06-15
featbit: new experimentation (release decision) module

cosmic-flood:

PR #955, opened 2026-08-11
🧹 refactor: remove data-analytics module

deleteLater:
Part of #921

removes the legacy data-analytics module and current experiment implementation

**MongoDB**

```javascript
for (const name of ["Experiments", "ExperimentMetrics"]) {
  if (db.getCollectionNames().includes(name)) {
    db.getCollection(name).drop();
  }
}
```

**Postgres**

```sql
DROP TABLE IF EXISTS experiments;
DROP TABLE IF EXISTS experiment_metrics;
```


cosmos-explorer (approved):
LGTM

PR #956, opened 2026-08-11
🧹 refactor: use new react front-end by default

cosmos-explorer:
- Rename front-end to front-end-v1
- Rename front-end-v2 to front-end
- Updated all related docs
- removed ui-change-validations.yml as it is for old front end


deleteLater (approved):
LGTM

